PR #5, opened 2018-08-22
add pkg.module and make pkg.main extension-agnostic

Rich-Harris:
Hi! This is just a small change that makes it easier to import worker-dom in a project using Rollup — the rollup-plugin-node-resolve plugin first looks for the `"module"` field in package.json, and only then falls back to `"main"`.

It also removes the extension from `"main"`, per the current Node recommendation, so that all bases are covered.

Thanks for this extraordinary project!


kristoferbaxter:
Thank you @Rich-Harris, please reach out if you run into trouble experimenting with it.

PR #12, opened 2018-08-27
Clarify usage instructions in the README

JamesMilnerUK:
When looking through the `Usage` section of the README I thought that there might be some room to clarify it slightly. 

* Preface that there are two variants at the beginning
* Clarify the example uses a direct script tag approach, and make using it as such more apparent by using a explanatory path 
* Separate the code for module / global variants

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,22 +19,31 @@ npm install @ampproject/worker-dom
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


JamesMilnerUK:
I signed it!

kristoferbaxter:
Looks great. Will merge.

PR #14, opened 2018-08-27
Fix paths for demos

### LINKED ISSUE #13: Library paths for demos are not pointing in the right place
Firstly thanks for this great library, top work! I just wanted to point out, the demos provided in the repository point at `./dist/index.mjs` and `./dist/index.js` respectively. The dist folder is actually the folder above, so to get any of the demos to work you have to change the URL to be `./../dist/index.js` or the module equivalent. 

Would you like me to raise a PR to fix this?

JamesMilnerUK:
As it stands the paths referenced in the demos are pointing in the wrong place, and need to be pointing at the dist folder in the directory above. As such none of the demos work out of the box. This PR fixes this issue.

Resolves #13

### DIFF
--- a/demo/dbmon.html
+++ b/demo/dbmon.html
@@ -5,28 +5,28 @@
   <title>DBMON</title>
   <meta name="viewport" content="width=device-width, initial-scale=1">
   <link href="dbmon.css" rel="stylesheet" type="text/css">
-  <script src="./dist/index.mjs" type="module"></script>
-  <script src="./dist/index.js" nomodule defer></script>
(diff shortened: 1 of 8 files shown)


kristoferbaxter:
Trying to fix this issue with a slightly different path in #20. Do you mind taking a look at that branch to see if you can reproduce the issue?

I am unable to reproduce on both `master` and `demo-refactor` branches locally.

JamesMilnerUK:
Hey @kristoferbaxter let me take a look. Thanks for looking into this :)

kristoferbaxter:
Closing this PR, since #20 addresses it.

PR #15, opened 2018-08-27
Add subheadings to demo page to stratify out different demo types

JamesMilnerUK:
This PR stratifies out the demos into sub categories making it easier for developers to find what they are looking for.

screenshot_20180827_183533


kristoferbaxter:
Good idea. Should make browsing the demos a bit more organized.

Issue #16, opened 2018-08-27
Debugger demo throws an ReferenceError on load


JamesMilnerUK:
Unfortunately the debugger demo currently throws an error upon loading the page:

```
ReferenceError: workerDocument is not defined
```
Let me know if I can provide any further assistance.

PR #20, opened 2018-08-29
Demo Refactor

kristoferbaxter:
Ping @JamesMilnerUK. Intended to address #14, #16, and #13. 

This change...
1. Moves every demo inside to a subdirectory inside `demo`s.
2. Changes the format of the demos to ensure consistency
    - All upgraded elements have the same identifier independent of AMP use-cases.
    - Resources needed for the specific demo live within the subdirectory
    - Resources from `worker-dom` are referenced from the `/dist` directory from the project root.
3. Identifies the DBMon implementation is driven by Preact.
4. Changes the command to run the demo webserver to `npm run demo`.
5. Adds additional comments to `README` explaining the webserver hosting demo code.
6. Additional comments added to debugging page explaining how to use it.

In #14, a different solution was provided to ensure access to the distributed version of the runtime and worker thread code. This change attempts to solve the problem in a different direction, relying on `server.mjs` to ensure the `dist` folder for `worker-dom` is exposed to the webserver. (https://github.com/ampproject/worker-dom/blob/master/demo/server.mjs#L24)


JamesMilnerUK:
@kristoferbaxter looking good and all working very nicely, and certainly resolves #14 and #16. I'd have a few small points that might improve the developer experience here:

- Placing a heads up in the README about the `npm run debug` command with a brief explanation that it starts up a small web server that runs the demos.
- Maybe `npm run debug` could be renamed to `npm run demos` to be more explicit ? 
- For the debugger demo, perhaps a brief explanation that the developer can open the developer console to experiment with the `workerDocument` and associated API?

kristoferbaxter:
Good suggestions. I've addressed all three items.

JamesMilnerUK:
@kristoferbaxter apologies as I missed the README already had a section for the demo server. Either way I think it's more clear now with your additional changes :+1:

PR #44, opened 2018-09-01
Ensure author script and document are on the same hostname

kristoferbaxter:
Author script and the document should reside on the same hostname to prevent bypassing CSP and leaking contents to other domains.


src/main-thread/worker.ts
```diff
@@ -18,48 +18,82 @@ import { MessageToWorker } from '../transfer/Messages';
...
 export function messageToWorker(worker: Worker, message: MessageToWorker) {
   worker.postMessage(message);
 }
+
+/**
+ * Validates the hostname of the author script matches the document.
+ * @param authorScriptURL 
+ * @returns boolean if the hostname of the document and the author script match.
+ */
+function authorSharesDocumentHostname(authorScriptURL: string): boolean {
```
choumx: We need a configuration API so the AMP can customize this for CDN URLs. E.g. `nytimes-com.cdn.ampproject.org` -> `nytimes.com`. 

I need to work on this anyways for custom sanitization rules for DOMPurify. Any preference for how this works?

kristoferbaxter: Quick thought: Perhaps we should model this after CSP? CSP format is already well known.

Doing a bit of research and will post a more thorough response.

PR #66, opened 2018-09-17
fix: change `npm run debug` to `npm run demo`

wupengFEX:
According to the readme description, you need to provide the `npm run demo` instead of `npm run debug`.


wupengFEX:
I signed it!

PR #67, opened 2018-09-17
fix: change `npm run debug` to `npm run demo`

wupengFEX:
According to the readme description, you need to provide the `npm run demo` instead of `npm run debug`.


JamesMilnerUK:
Confusing as I thought "debug" was already moved to "demo" in the demo refactoring PR: https://github.com/ampproject/worker-dom/blob/d437a947ed6ad31b7af69d35fb675ea2636d3769/package.json . Either way, yes looks like this needs updating to match the README

Issue #110, opened 2018-10-12
FR: Support unmangled "debug" output


choumx:
I came across this when debugging #109. Would be easier to debug issues without obfuscated function names etc.

choumx:
Typically I see it flipped: `index.js` is unminified, `index.min.js` is. Up to you though.

PR #112, opened 2018-10-12
Add unminified builds

kristoferbaxter:
Addresses #110.

Chose the format `unminified.${filename}.${extension}` so the default would still use the minification passes that are custom to the closure compiler version in the library.


choumx:
Oops, merged when I actually meant to approve. Sorry!

kristoferbaxter:
No worries! I'll release the version on NPM now.

PR #113, opened 2018-10-17
Support configurable hooks for DOMPurifySanitizer

choumx:
TODO: Add unit tests. :)

### DIFF
--- a/src/main-thread/DOMPurifySanitizer.ts
+++ b/src/main-thread/DOMPurifySanitizer.ts
@@ -18,12 +18,70 @@ import purify from 'dompurify';
 
 const propertyToAttribute: { [key: string]: string } = {}; // TODO(choumx): Fill this in.
 
+/**
+ * Object containing optional callbacks. Use this to configure
+ * DOMPurify hooks before/after sanitization.
+ * @see https://github.com/cure53/DOMPurify#hooks
+ */
+interface DOMPurifySanitizerCallbacks {
+  afterSanitize?: (purify: DOMPurify) => void;
+  beforeSanitize?: (purify: DOMPurify) => void;
+  nodeWasRemoved?: (node: Node) => void;
+}
+
 export class DOMPurifySanitizer implements Sanitizer {
+  callbacks_: DOMPurifySanitizerCallbacks;
+  config_: Object;
+  wrapper_: HTMLElement;
+
+  constructor() {
+    this.config_ = {};
+    this.wrapper_ = document.createElement('div');
+  }
+
+  /**
+   * @param config https://github.com/cure53/DOMPurify#can-i-configure-it
+   * @param callbacks
+   */
+  configure(config: Object, callbacks: DOMPurifySanitizerCallbacks) {
+    this.config_ = config;
+    this.callbacks_ = callbacks;
+  }
+
   /**
    * @param node
+   * @return False if node was removed during sanitization. Otherwise, true.
    */
-  sanitize(node: Node): void {
-    purify.sanitize(node, { IN_PLACE: true });
+  sanitize(node: Node): boolean {
+    if (this.callbacks_ && this.callbacks_.beforeSanitize) {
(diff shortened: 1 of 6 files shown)


src/main-thread/mutator.ts
```diff
@@ -107,7 +115,7 @@ export function mutate(
   // }
   // this.lastGestureTime = lastGestureTime;
   stringValues.forEach(value => storeString(value));
-  nodes.forEach(node => createNode(node));
+  nodes.forEach(node => createNode(node, sanitizer));
```
choumx: @kristoferbaxter Security hole! Also, I don't really understand why both this line and line 51 are necessary. Would you mind explaining?

src/main-thread/DOMPurifySanitizer.ts
```diff
@@ -18,12 +18,70 @@ import purify from 'dompurify';
...
+    }
+    // DOMPurify sanitizes unsafe nodes by detaching them from parents.
+    // So, if `node` itself is unsafe and has no parent: runtime error.
+    // To avoid this, wrap `node` in a div if it has no parent.
+    const useWrapper = !node.parentNode;
+    if (useWrapper) {
+      this.wrapper_.appendChild(node);
+    }
+    const parent = node.parentNode || this.wrapper_;
+    purify.sanitize(parent, Object.assign({}, this.config_, { IN_PLACE: true }));
```
kristoferbaxter: This `Object.assign` looks static, can we just perform the assignment in the constructor?

src/main-thread/DOMPurifySanitizer.ts
```diff
@@ -18,12 +18,70 @@ import purify from 'dompurify';
...
+    this.config_ = {};
+    this.wrapper_ = document.createElement('div');
+  }
+
+  /**
+   * @param config https://github.com/cure53/DOMPurify#can-i-configure-it
+   * @param callbacks
+   */
+  configure(config: Object, callbacks: DOMPurifySanitizerCallbacks) {
+    this.config_ = config;
```
kristoferbaxter:
```suggestion
    this.config_ = Object.assign({}, config, { IN_PLACE: true });
```

src/main-thread/mutator.ts
kristoferbaxter: During a mutation a new node may be transferred that didn't previously exist. These nodes are immediately created so when a mutation attaching to them or attaching them to the document executes the node exists. Otherwise the representation used in line 51 fails.

src/main-thread/nodes.ts
```diff
@@ -48,9 +48,9 @@ export function createNode(skeleton: TransferrableNode): RenderableElement {
 
   const namespace: string | undefined =
     skeleton[TransferrableKeys.namespaceURI] !== undefined ? getString(skeleton[TransferrableKeys.namespaceURI] as number) : undefined;
-  const node: HTMLElement | SVGElement = namespace
-    ? (document.createElementNS(namespace, getString(skeleton[TransferrableKeys.nodeName])) as SVGElement)
-    : document.createElement(getString(skeleton[TransferrableKeys.nodeName]));
+  const nodeName = getString(skeleton[TransferrableKeys.nodeName]);
+  let node: HTMLElement | SVGElement = namespace ? (document.createElementNS(namespace, nodeName) as SVGElement) : document.createElement(nodeName);
```
kristoferbaxter: `node` doesn't appear to be reassigned, can this be a `const`? I wonder why the linter didn't catch this.

```suggestion
  const node: HTMLElement | SVGElement = namespace ? (document.createElementNS(namespace, nodeName) as SVGElement) : document.createElement(nodeName);
```

choumx:
Suggested changes are cool. :) 

I'm going to split the unit tests into a separate PR. Cool if we merge this first while we sort out main-thread testing?

kristoferbaxter:
Makes sense, will merge and release.

kristoferbaxter:
Released as `0.1.4`.

Issue #111, opened 2018-10-12
Implement `querySelector` and `querySelectorAll`


kristoferbaxter:
Used and beloved by many web developers, and completely missing from `worker-dom`.

Let's add it in.

This description is intentionally small to allow @nainar to fill in as details rise in the implementation.

nainar:
Logging current implementation plan per discussion:
- [ ] Implement querySelectorAll for the following:
  - [ ] class selector
  - [ ] ID selector
  - [ ] attribute selector
  - [ ] tag selector
- [ ] Implement querySelector to return the 0th item result of calling querySelector
- [ ] Wait for use case to add support for more complex selectors

kristoferbaxter:
We'll revisit adding more complex selector support based on requested use cases.

PR #115, opened 2018-10-18
Adds support for ID, Class & Tag selectors to Element.querySelector

nainar:
Addresses #111

### DIFF
--- a/src/worker-thread/dom/Element.ts
+++ b/src/worker-thread/dom/Element.ts
@@ -386,5 +386,31 @@ export class Element extends Node {
   public getElementsByTagName(tagName: string): Element[] {
     return matchChildrenElements(this, tagName === '*' ? _ => true : element => element.tagName === tagName);
   }
+  /**
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector
+   * @return Element with matching selector.
+   */
+  public querySelector(selector: string): Element | null {
+    let selectors = selector.split(' ');
+    // Shortcut for single id, class, tag selectors
+    if (selectors.length === 1) {
+      if (selector[0] == '#') {
+        return this.ownerDocument.getElementById(selector.substr(1));
+      } else if (selector[0] == '.') {
+        const matches = this.getElementsByClassName(selector.substr(1));
+        if (matches) {
+          return matches[0];
+        }
+      } else {
+        selector = selector.toLowerCase();
+        const matches = this.getElementsByTagName(selector);
+        if (matches) {
+          return matches[0];
+        }
+      }
+    }
+    //TODO(nainar): More complex selectors
+    return null;
+  }
 }
 reflectProperties([{ id: [''] }], Element);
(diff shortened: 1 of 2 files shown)


nainar:
@kristoferbaxter PTAL? Can't assign you as a reviewer.

src/worker-thread/dom/Element.ts
```diff
@@ -386,5 +386,28 @@ export class Element extends Node {
   public getElementsByTagName(tagName: string): Element[] {
     return matchChildrenElements(this, tagName === '*' ? _ => true : element => element.tagName === tagName);
   }
+
+  /**
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector
+   * @return Element with matching selector.
```
kristoferbaxter: Can you add a JS Doc comment for the argument `selector`?

src/worker-thread/dom/Element.ts
```diff
@@ -386,5 +386,28 @@ export class Element extends Node {
...
+   */
+  public querySelector(selector: string): Element | null {
+    if (selector[0] == '#') {
```
kristoferbaxter: Is there a reason for loose equality checking here?

src/worker-thread/dom/Element.ts
```diff
@@ -386,5 +386,28 @@ export class Element extends Node {
...
+      return this.ownerDocument.getElementById(selector.substr(1));
+    } else if (selector[0] == '.') {
+      const matches = this.getElementsByClassName(selector.substr(1));
```
kristoferbaxter: According to MDN `querySelector` uses the entire document's elements to find matches, then limits to the scope of the element requested from. Are we sure the behaviour is the same here if we intentionally limit to the `baseElement` the query selection was requested from?

[MDN Document](https://developer.mozilla.org/en-US/docs/Web/API/Element/querySelector)
> The entire hierarchy of elements is considered when matching, including those outside the set of elements including baseElement and its descendants

Just asking for my own knowledge, and perhaps to push for a comment once we have an explanation.

src/worker-thread/dom/Element.ts
nainar: Done.

src/worker-thread/dom/Element.ts
```diff
@@ -386,5 +386,28 @@ export class Element extends Node {
...
+      if (matches) {
+        return matches[0];
+      }
+    } else {
+      selector = selector.toLowerCase();
+      const matches = this.getElementsByTagName(selector);
```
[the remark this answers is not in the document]
nainar: Done.

src/worker-thread/dom/Element.ts
nainar: Nope. My mistake. Fixed.

src/worker-thread/dom/Element.ts
nainar: Fixed this to first check the document and then filter the results.

kristoferbaxter:
@nainar – Changes from feedback look great! Want to merge this as is and add further selector support in a followup PR?

nainar:
Yup, that's the intention :)

PR #116, opened 2018-10-18
Fix Node::contains order of operations. 

nainar:
Currently in Node:::contains() if we return the status of checking if our children contain `otherNode` we are looking for, ignoring that we may ourselves be the node that is being searched for due to the return.

We could fix this by either caching the result of checking our children and returning the union of checking if the node itself or it's children contain the node. OR (implemented method) first check ourself to see if we are `otherNode`. The latter is implemented as it saves us checking our children nodes. 

A test has also been added.

### DIFF
--- a/src/test/node/contains.ts
+++ b/src/test/node/contains.ts
@@ -15,14 +15,13 @@
  */
 
 import test from 'ava';
-import { NodeType } from '../../worker-thread/dom/Node';
-import { Element } from '../../worker-thread/dom/Element';
+import { documentForTesting as document } from '../../worker-thread/dom/Document';
 
 test.beforeEach(t => {
   t.context = {
-    node: new Element(NodeType.ELEMENT_NODE, 'div', null),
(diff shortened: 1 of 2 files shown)


nainar:
@kristoferbaxter for review

src/test/node/contains.ts
```diff
@@ -15,14 +15,13 @@
...
 test.beforeEach(t => {
   t.context = {
-    node: new Element(NodeType.ELEMENT_NODE, 'div', null),
-    child: new Element(NodeType.ELEMENT_NODE, 'div', null),
-    childTwo: new Element(NodeType.ELEMENT_NODE, 'div', null),
+    node: document.createElement('div'),
+    child: document.createElement('div'),
+    childTwo: document.createElement('div'),
   };
 });
```
kristoferbaxter: If we're switching to a `beforeEach` with a `document`, we should likely remove all child nodes of the document in the `afterEach` lifecycle.

nainar: Done.

PR #117, opened 2018-10-18
API to specify author URL separate from base element

choumx:
https://github.com/ampproject/amphtml/pull/18665/files#r224279646

Open to better names than `upgrade()`. :)


choumx:
Friendly ping. :)

PR #121, opened 2018-10-20
Add Support for Comment Nodes

kristoferbaxter:
Several popular frameworks use `Comment` nodes to reserve spots in DOM for future mutations.

This PR is WIP adding support for Comment nodes.

### DIFF
--- a/demo/comments/comments.js
+++ b/demo/comments/comments.js
@@ -1,8 +1,6 @@
 const comment = document.createComment('Test');
 const el = document.createElement('div');
-const replacer = document.createElement('p');
 
-replacer.appendChild(document.createTextNode('replacer'));
 el.appendChild(document.createTextNode('bar'));
 el.classList.add('foo');
 document.body.appendChild(comment);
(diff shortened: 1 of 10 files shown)


demo/comments/comments.js
```diff
@@ -11,7 +9,13 @@ document.body.appendChild(el);
 let replaced = false;
 el.addEventListener('click', function() {
   if (!replaced) {
+    const replacer = document.createElement('p');
```
kristoferbaxter: This change is to avoid an issue that needs fixing in a separate PR (#125).

src/test/mutationobserver/replaceChild.ts
```diff
@@ -41,6 +42,60 @@ test.cb.serial('replaceChild mutation, only node', t => {
   document.body.replaceChild(p, div);
 });
 
+test.cb.serial('replaceChild mutation, replace first with second', t => {
```
kristoferbaxter: Increased test coverage here since this functionality was silently broken for many cases.

src/test/comment/textContent.ts
```diff
@@ -0,0 +1,52 @@
...
+
+test('get textContent', t => {
+  const { comment } = t.context as { comment: Comment };
+
+  t.is(comment.textContent, 'default value');
+});
+
+test('set textContent', t => {
+  const { comment } = t.context as { comment: Comment };
+
```
nainar: Add a test to assert that it is `default value` before
```suggestion

t.is(comment.textContent, 'default value');
```

PR #126, opened 2018-10-23
Adds support for ID, Class & Tag selectors to Element.querySelectorAll

nainar:
Addresses #111

### DIFF
--- a/src/test/htmlelement/querySelectorAll.ts
+++ b/src/test/htmlelement/querySelectorAll.ts
@@ -66,20 +66,28 @@ test('test Element.querySelectorAll is case insensitive with regards to tags', t
 });
 
 test('test Element.querySelector returns the first result of Element.querySelectorAll', t => {
-  const { parentDiv } = t.context as { parentDiv: HTMLElement; div: HTMLElement };
+  const { parentDiv } = t.context as { parentDiv: HTMLElement };
   let querySelectorAllResults = document.querySelectorAll('div');
   t.not(querySelectorAllResults, null);
-  t.deepEqual(querySelectorAllResults![0], document.querySelector('div'));
+  if (querySelectorAllResults) {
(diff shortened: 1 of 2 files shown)


nainar:
@kristoferbaxter this is a copy of Element::querySelector, except for the return case.

kristoferbaxter:
I'm curious if `querySelector` could be refactored to delegate functionality to `querySelectorAll` and just return the 0th value.

This would reduce the amount of code significantly. Thoughts?

src/test/htmlelement/querySelectorAll.ts
```diff
@@ -0,0 +1,85 @@
...
+  t.deepEqual(parentDiv.querySelectorAll('DIV'), [div]);
+
+  t.deepEqual(document.querySelectorAll('div'), document.querySelectorAll('DIV'));
+  t.deepEqual(parentDiv.querySelectorAll('div'), parentDiv.querySelectorAll('DIV'));
+});
+
+test('test Element.querySelector returns the first result of Element.querySelectorAll', t => {
+  const { parentDiv } = t.context as { parentDiv: HTMLElement; div: HTMLElement };
+  let querySelectorAllResults = document.querySelectorAll('div');
+  t.true(!!querySelectorAllResults);
```
nainar: Is there a better way for checking that this isn't null?

src/test/htmlelement/querySelectorAll.ts
```diff
@@ -0,0 +1,85 @@
...
+  t.deepEqual(querySelectorAllResults![0], document.querySelector('div'));
```
nainar: Is there a safer way of telling typescript that I have already asserted before that this thing isn't null?

kristoferbaxter: You could wrap this test in a conditional, since you're certain it will pass the test will always run.

PR #129, opened 2018-10-25
Callbacks for worker message debugging

choumx:
Adds support for users to configure `onSendMessage(data)` and `onReceiveMessage(data)` callbacks, where `data` is a human-readable copy of the raw `postMessage` data.

### Before:
image

### After:
image

### Notes
- Increases the main-thread binary size by ~0.7KB. One option is to only support this in the "safe" output variant.
- I originally tried using non-const enums so that TS would auto-generate [reverse mappings](https://www.typescriptlang.org/docs/handbook/enums.html#enums-at-runtime), but that approach turned out to be less readable and less performant due to lack of enum value inlining. According to TS team, `--preserveConstEnums` apparently is not meant for this use case either.
- Everything lives in main-thread/debugging.ts instead of colocated with messages due to use of `getNode()` and other main-thread code.


choumx:
> Changes for moving debugging out of index and only into index.safe.

Done. Reverted the bundle-size bump for index.mjs.

```
 PASS  ./dist/index.mjs: 1.7KB < maxSize 2KB (brotli) 
 PASS  ./dist/index.safe.mjs: 7.25KB < maxSize 8KB (brotli) 
```

choumx:
@kristoferbaxter Addressed comments, I'll merge tonight or tomorrow morning unless you object. :)

PR #145, opened 2018-11-01
CSSStyleDeclaration values can be zero

### LINKED ISSUE #141: Can't set value "0" for CSS style
Example of code that doesn't work:
`element.style.top = 0`

Below is checking value for truthy, which causes it to ignore 0:
https://github.com/ampproject/worker-dom/blob/master/src/worker-thread/css/CSSStyleDeclaration.ts#L108

kristoferbaxter:
Fixes #141.

### DIFF
--- a/src/worker-thread/css/CSSStyleDeclaration.ts
+++ b/src/worker-thread/css/CSSStyleDeclaration.ts
@@ -105,7 +105,7 @@ export class CSSStyleDeclaration implements StyleDeclaration {
    * @return value stored for the provided key.
    */
   public getPropertyValue(key: string): string {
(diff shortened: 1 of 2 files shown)


src/worker-thread/css/CSSStyleDeclaration.ts
```diff
@@ -105,7 +105,7 @@ export class CSSStyleDeclaration implements StyleDeclaration {
    * @return value stored for the provided key.
    */
   public getPropertyValue(key: string): string {
```
prateekbh: we are returning `number`, but return type is `string`? A comment may be

Issue #122, opened 2018-10-22
Better DOM property support


choumx:
A few issues I've noticed:
- [x] Boolean and enumerated attributes are not handled correctly (setting attribute value to `String(property)` is incorrect). Also need main-thread support for attribute removal (important for boolean attributes).
- [x] Some important properties are not implemented e.g. `HTMLInputElement.value`. Forwarding attribute mutations are insufficient for some of these properties.

One repro case:
```js
const input = document.createElement('input');
input.setAttribute('type', 'text');
document.body.appendChild(input);

const btn = document.createElement('button');
btn.textContent = 'Clear input and toggle disabled'
btn.addEventListener('click', () => {
  input.value = '';
  input.disabled = !input.disabled;
});
document.body.appendChild(btn);
```

```html
<input type="text"><button>prompt</button>
```

PR #153, opened 2018-11-05
Support HTMLInputElement 'value' and 'checked' properties

choumx:
Partial for #122. 

- Add `value`, `valueAsNumber`, `valueAsDate`, `checked` props for `HTMLInputElement`.
- Forward property mutations for these from worker to page.

TODO: Support SYNC events for properties other than `value`, e.g. `checked`.


choumx:
Merging. Happy to address any feedback in a follow-up.

PR #169, opened 2018-11-15
Node.cloneNode

kristoferbaxter:
Adds support for `Node.cloneNode`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "dompurify": "1.0.8"
   },
   "devDependencies": {
-    "@ampproject/rollup-plugin-closure-compiler": "0.8.3",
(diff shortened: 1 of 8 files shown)


kristoferbaxter:
Note: We'll need to verify the `_index_` varies between the clone and source with tests. Also, we should write tests verifying the event handlers are not registered on the clone for future implementors.

kristoferbaxter:
This PR likely should wait for the change to the closure compiler plugin that gives us back a few % in brotli compressed size of the resources.

westonruter:
Something's not right here, as `Element.cloneNode()` is constructing `HTMLUnknownElement` instances in the host page: #614.

PR #224, opened 2019-01-10
Upgrade Ava to 1.0

kristoferbaxter:
This upgrade was surprisingly complicated – 1.0 added the ability to type `t.context`, which meant all usages across each test needed to be upgraded.

Thankfully, this gave the opportunity to move all tests over to `document` constructors, and remove the restriction around a singleton document.

### DIFF
--- a/src/worker-thread/dom/Comment.ts
+++ b/src/worker-thread/dom/Comment.ts
@@ -19,11 +19,12 @@ import { NumericBoolean } from '../../utils';
 import { TransferrableKeys } from '../../transfer/TransferrableKeys';
 import { NodeType } from '../../transfer/TransferrableNodes';
 import { store as storeString } from '../strings';
+import { Node } from './Node';
 
 // @see https://developer.mozilla.org/en-US/docs/Web/API/Comment
 export class Comment extends CharacterData {
-  constructor(data: string) {
-    super(data, NodeType.COMMENT_NODE, '#comment');
+  constructor(data: string, ownerDocument: Node) {
+    super(data, NodeType.COMMENT_NODE, '#comment', ownerDocument);
     this[TransferrableKeys.creationFormat] = {
       [TransferrableKeys.index]: this[TransferrableKeys.index],
       [TransferrableKeys.transferred]: NumericBoolean.FALSE,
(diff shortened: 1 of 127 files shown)


src/worker-thread/dom/Document.ts
```diff
@@ -19,6 +19,7 @@ import { HTMLElement } from './HTMLElement';
 import './HTMLAnchorElement';
 import './HTMLButtonElement';
 import './HTMLDataElement';
+import './HTMLDataListElement';
```
kristoferbaxter: Note, the change to tests now running all Element creation thru `document` yielded a bug find!

`HTMLDataListElement` was not available to users, and was falling back to `HTMLElement`.

src/worker-thread/dom/Comment.ts
```diff
@@ -19,11 +19,12 @@ import { NumericBoolean } from '../../utils';
...
 import { NodeType } from '../../transfer/TransferrableNodes';
 import { store as storeString } from '../strings';
+import { Node } from './Node';
 
 // @see https://developer.mozilla.org/en-US/docs/Web/API/Comment
 export class Comment extends CharacterData {
-  constructor(data: string) {
-    super(data, NodeType.COMMENT_NODE, '#comment');
+  constructor(data: string, ownerDocument: Node) {
+    super(data, NodeType.COMMENT_NODE, '#comment', ownerDocument);
```
kristoferbaxter: We now pipe the `ownerDocument` to each Node during construction, and avoid the singleton `globalDocument` pattern used before.

src/worker-thread/dom/HTMLMapElement.ts
```diff
@@ -29,7 +29,7 @@ export class HTMLMapElement extends HTMLElement {
     return matchChildrenElements(this as Element, element => element.tagName === 'area');
   }
 }
-registerSubclass('link', HTMLMapElement);
+registerSubclass('map', HTMLMapElement);
```
kristoferbaxter: Another find from changing the tests to use a `document` to construct all `Element`s. Both `map` and `link` elements were not correctly configured.

PR #234, opened 2019-01-17
Allow delegating of script fetching to caller

choumx:
This allows AMP to perform the fetch of worker/author scripts which:

1. Avoids need for a fetch polyfill.
2. Allows AMP to set its typical request headers and query params.
3. Enforce a size constraint on the author script.

/to @kristoferbaxter

### DIFF
--- a/src/main-thread/index.ts
+++ b/src/main-thread/index.ts
@@ -14,15 +14,11 @@
  * limitations under the License.
  */
 
-import { fetchAndInstall, install } from './install';
+import { fetchAndInstall } from './install';
 
 export function upgradeElement(baseElement: Element, workerDOMUrl: string): void {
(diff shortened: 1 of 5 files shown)


src/main-thread/index.ts
```diff
@@ -14,15 +14,11 @@
...
   const authorURL = baseElement.getAttribute('src');
   if (authorURL) {
-    upgrade(baseElement, authorURL, workerDOMUrl);
+    fetchAndInstall(baseElement as HTMLElement, authorURL, workerDOMUrl);
   }
 }
-
-export function upgrade(baseElement: Element, authorURL: string, workerDOMUrl: string): void {
-  install(baseElement as HTMLElement, authorURL, workerDOMUrl);
-}
```
choumx: Removed to minimize API surface.

Issue #232, opened 2019-01-16
Infinite recursion at CSSStyleDeclaration.getPropertyValue


choumx:
Only reproducible at current HEAD (I think starting at #185). Not reproducible on 0.2.8. 

A:
https://github.com/ampproject/worker-dom/blob/21dc0ff12ac62c9cbb3b7d50ed901d4eb07beaa4/src/worker-thread/css/CSSStyleDeclaration.ts#L68-L70

B:
https://github.com/ampproject/worker-dom/blob/21dc0ff12ac62c9cbb3b7d50ed901d4eb07beaa4/src/worker-thread/css/CSSStyleDeclaration.ts#L108-L110

A -> B -> A etc.

Results in error:
```
Uncaught RangeError: Maximum call stack size exceeded
    at CSSStyleDeclaration.getPropertyValue (...)
    at CSSStyleDeclaration.get (...)
    at CSSStyleDeclaration.getPropertyValue (...)
    at CSSStyleDeclaration.get (...)
    ...
```

Reproduced from author JS:
```js
document.getElementById('hello').addEventListener('click', ()=>{
  const el = document.createElement('h1'); // Infinite recursion starts here on click.
  el.textContent = 'Hello World!';
  document.body.appendChild(el);
});
```

Stack trace:
```
getPropertyValue (/examples/amp-script/hello-world.js:formatted:1600)
get (/examples/amp-script/hello-world.js:formatted:1557)
setProperty (/examples/amp-script/hello-world.js:formatted:1627)
set (/examples/amp-script/hello-world.js:formatted:1560)
CSSStyleDeclaration (/examples/amp-script/hello-world.js:formatted:1571)
Element (/examples/amp-script/hello-world.js:formatted:1821)
HTMLElement (/examples/amp-script/hello-world.js:formatted:2226)
createElementNS (/examples/amp-script/hello-world.js:formatted:4610)
createElement (/examples/amp-script/hello-world.js:formatted:4605)
document.getElementById.addEventListener (/examples/amp-script/hello-world.js:formatted:4892)
dispatchEvent (/examples/amp-script/hello-world.js:formatted:815)
(anonymous) (/examples/amp-script/hello-world.js:formatted:4438)
```

kristoferbaxter:
Good find! We should add a test to prevent a regression with the fix.

choumx:
[...]

PR #238, opened 2019-01-18
Filter invalid styles from CSSStyleDeclaration

kristoferbaxter:
- [x] Fix recursion
- [x] Tests

Addresses #232

### DIFF
--- a/src/main-thread/worker.ts
+++ b/src/main-thread/worker.ts
@@ -44,9 +44,7 @@ export function createWorker(
   const keys: Array<string> = [];
   const hydratedNode = createHydrateableNode(baseElement);
   for (const key in document.body.style) {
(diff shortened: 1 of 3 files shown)


src/main-thread/worker.ts
```diff
@@ -44,7 +44,9 @@ export function createWorker(
   const keys: Array<string> = [];
   const hydratedNode = createHydrateableNode(baseElement);
   for (const key in document.body.style) {
-    keys.push(`'${key}'`);
+    if (isNaN(key as any)) {
```
choumx: Nice use of `isNaN`. Maybe we should do this upstream in `appendKeys()` instead?

kristoferbaxter: Addressed with 08e1449.

PR #252, opened 2019-01-25
Implement debugging hook for consumeInitialDOM

choumx:
Allows callers to inspect/print the `HydrateableNode` that used during worker bootstrapping.


kristoferbaxter:
@choumx – There is a conflict now with merged changes from Justin's work. Can you take a look and we can re-review?

choumx:
On it. Not trivial to rebase but I have an idea of how to do it.

choumx:
Done, PTAL. Instead of deserializing the HydrateableNode, I create another "readable" version from the `baseElement`.

kristoferbaxter:
Looking now.

src/main-thread/worker.ts
```diff
@@ -76,7 +76,11 @@ export function createWorker(
       ${authorScript}
     }).call(WorkerThread.workerDOM);
 //# sourceURL=${encodeURI(authorScriptURL)}`;
-  return new Worker(URL.createObjectURL(new Blob([code])));
+  const worker = new Worker(URL.createObjectURL(new Blob([code])));
+  if (callbacks && callbacks.onCreateWorker) {
```
kristoferbaxter: Here's another place where the debug build varies from the production build. Let's make sure these additions are only present in the debug build (in a separate PR).

PR #274, opened 2019-02-05
Terminate Worker When Threshold for Interaction is Surpassed

kristoferbaxter:
Adds a threshold for bound user interaction events leading to mutations when a Sanitizer is applied to mutations.


kristoferbaxter:
Mutation pump handles this case now for AMP.

PR #284, opened 2019-02-13
🐛Change the return type of `getNode` to be nullable.

### LINKED ISSUE #280: src/main-thread/nodes.ts#getNode() should have nullable return type
Given a bogus ID, the main-thread's `getNode()` can return `null`. Therefore, its return type should be nullable i.e. `RenderableElement | null`. 

This will require downstream changes. 

Alternative: Retain the non-nullable return type, but throw a runtime exception if the `id` can't be found.

/to @delima02 /cc @kristoferbaxter

delima02:
* Change the return type of `src/main-thread/nodes.ts#getNode()` to be nullable.
* In places where this method is called, expect a nullable return value.
* Fixes #280

🐛
@choumx @jridgewell

### DIFF
--- a/src/main-thread/commands/bounding-client-rect.ts
+++ b/src/main-thread/commands/bounding-client-rect.ts
@@ -30,6 +30,10 @@ import { NumericBoolean } from '../../utils';
 export function process(worker: Worker, mutation: TransferrableMutationRecord): void {
   const target = getNode(mutation[TransferrableKeys.target]);
 
+  if (!target) {
+    return;
+  }
+
   const boundingRect = target.getBoundingClientRect();
   messageToWorker(worker, {
     [TransferrableKeys.type]: MessageType.GET_BOUNDING_CLIENT_RECT,
--- a/src/main-thread/commands/bounding-client-rect.ts
+++ b/src/main-thread/commands/bounding-client-rect.ts
@@ -28,10 +28,11 @@ import { NumericBoolean } from '../../utils';
  * @param mutation mutation record containing commands to execute.
  */
 export function process(worker: Worker, mutation: TransferrableMutationRecord): void {
-  const target = getNode(mutation[TransferrableKeys.target]);
+  const nodeId = mutation[TransferrableKeys.target];
+  const target = getNode(nodeId);
 
   if (!target) {
-    console.error('ERROR: getNode() yields a null value. Node id was not found.');
+    console.error('getNode() yields a null value. Node id (' + nodeId + ') was not found.');
     return;
   }
 
--- a/src/main-thread/commands/bounding-client-rect.ts
+++ b/src/main-thread/commands/bounding-client-rect.ts
@@ -31,6 +31,7 @@ export function process(worker: Worker, mutation: TransferrableMutationRecord): 
   const target = getNode(mutation[TransferrableKeys.target]);
 
   if (!target) {
+    console.error('ERROR: getNode() yields a null value. Node id was not found.');
     return;
   }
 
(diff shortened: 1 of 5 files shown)


src/main-thread/nodes.ts
```diff
@@ -91,7 +91,7 @@ export function createNode(skeleton: TransferrableNode, sanitizer?: Sanitizer):
 /**
  * Returns the real DOM Element corresponding to a serialized Element object.
  * @param id
- * @return RenderableElement
+ * @return RenderableElement | null
```
choumx: This only changes documentation. In TypeScript, you want to change the function type: https://www.typescriptlang.org/docs/handbook/functions.html

Once you change it, the compiler should show several type errors caused by the change. Grab me for a chat when you get to that point.

src/main-thread/mutator.ts
```diff
@@ -42,7 +42,14 @@ const mutators: {
   [key: number]: (mutation: TransferrableMutationRecord, target: Node) => void;
 } = {
   [MutationRecordType.CHILD_LIST](mutation: TransferrableMutationRecord, target: HTMLElement) {
-    (mutation[TransferrableKeys.removedNodes] || []).forEach(node => getNode(node[TransferrableKeys.index]).remove());
+    (mutation[TransferrableKeys.removedNodes] || []).forEach(node => {
+      const retrieved = getNode(node[TransferrableKeys.index]);
```
choumx: s/retrieved/node

src/main-thread/mutator.ts
```diff
@@ -42,7 +42,14 @@ const mutators: {
...
+      if (!retrieved) {
+        console.error('ERROR: getNode() yields a null value. Node id was not found.');
```
choumx: "ERROR" is redundant. Also, include the missing ID in the message.

src/main-thread/commands/bounding-client-rect.ts
```diff
@@ -31,6 +31,7 @@ export function process(worker: Worker, mutation: TransferrableMutationRecord):
   const target = getNode(mutation[TransferrableKeys.target]);
 
   if (!target) {
+    console.error('ERROR: getNode() yields a null value. Node id was not found.');
```
choumx: Are `console.error` stripped from the production build?

src/main-thread/commands/event-subscription.ts
```diff
@@ -134,6 +134,7 @@ export function process(worker: Worker, mutation: TransferrableMutationRecord):
   const target = getNode(mutation[TransferrableKeys.target]);
 
   if (!target) {
+    console.error('ERROR: getNode() yields a null value. Node id was not found.');
```
choumx: There's a similar issue here with amphtml -- duplicate error strings can cause bundle size bloat. 

@kristoferbaxter Do you have a preference here?

src/main-thread/mutator.ts
delima02: Done!

src/main-thread/mutator.ts
delima02: I don't think I understand this comment. Can you give me a hand?

choumx: Replace "retrieved" with "node". It's a reference from `sed`: https://thoughtbot.com/blog/sed-102-replace-in-place

delima02: "node" already exists in this scope. Maybe "retrievedNode" or something specific?

choumx: I'd rename the existing `node` to `removed` or `nodeReference`.

choumx:
Did we check if `console.error` is stripped from the production bundle?

src/main-thread/commands/event-subscription.ts
kristoferbaxter: If we think the errors are important, I'll add a build step that moves them to codes and generates a json file with mappings for code -> string. We can host this like we're planning with AMPHTML.

choumx: They are stripped from the prod build but preserved on the dev build. 

I think this is good to go until we have an error code mapping system. Thanks Kris.

PR #286, opened 2019-02-13
🐛Implement navigator.

delima02:
- Add navigator with the userAgent property to worker-dom.
- Fixes #285 

@jridgewell @choumx

PR #288, opened 2019-02-14
Implement Navigator.userAgent property.

### LINKED ISSUE #285: Implement Navigator.userAgent.
Needed by #282 .

/to @delima02 /cc @jridgewell @choumx

delima02:
* Add navigator with the userAgent property to worker-dom.
* Fixes #285 

@jridgewell @choumx

### DIFF
--- a/src/main-thread/worker.ts
+++ b/src/main-thread/worker.ts
@@ -46,6 +46,7 @@ export function createWorker(
   for (const key in document.body.style) {
     keys.push(key);
   }
+  const userAgent = window.navigator.userAgent;
   const code = `
     'use strict';
     ${workerDOMScript}
@@ -56,6 +57,7 @@ export function createWorker(
       var localStorage = this.localStorage;
       var location = this.location;
       var defaultView = document.defaultView;
+      var navigator = window.navigator;
       var Node = defaultView.Node;
       var Text = defaultView.Text;
       var Element = defaultView.Element;
@@ -63,6 +65,7 @@ export function createWorker(
       var Document = defaultView.Document;
       var Event = defaultView.Event;
       var MutationObserver = defaultView.MutationObserver;
+      navigator.userAgent = ${JSON.stringify(userAgent)};
 
       function addEventListener(type, handler) {
         return document.addEventListener(type, handler);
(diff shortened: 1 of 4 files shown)


delima02:
Ignore #286, I branched off of the wrong branch. This is the correct one. @jridgewell

src/main-thread/worker.ts
```diff
@@ -56,13 +57,15 @@ export function createWorker(
...
       var defaultView = document.defaultView;
+      var navigator = window.navigator;
       var Node = defaultView.Node;
       var Text = defaultView.Text;
       var Element = defaultView.Element;
       var SVGElement = defaultView.SVGElement;
       var Document = defaultView.Document;
       var Event = defaultView.Event;
       var MutationObserver = defaultView.MutationObserver;
+      navigator.userAgent = ${JSON.stringify(userAgent)};
```
choumx: Why is `stringify` necessary?

src/worker-thread/index.ts
```diff
@@ -52,6 +52,7 @@ declare var __ALLOW_POST_MESSAGE__: boolean;
 const doc = createDocument(__ALLOW_POST_MESSAGE__ ? (self as DedicatedWorkerGlobalScope).postMessage : undefined);
 export const workerDOM: WorkerDOMGlobalScope = {
   document: doc,
+  navigator: { userAgent: '' },
```
choumx: Add a comment saying that `userAgent` is propagated at worker creation time.

choumx:
PR title is misleading -- we're only implementing `navigator.userAgent`, not all of `Navigator`.

src/main-thread/worker.ts
delima02: Without it, the entire `code` (outer) string gets messed up and an error is thrown. Using `stringify` just makes sure it correctly makes it as a string within a string.

choumx: > Without it, the entire `code` (outer) string gets messed up and an error is thrown.

There's a simpler fix here. Hint: what error does it throw and why?

delima02: The error thrown goes as follows:
`Uncaught SyntaxError: missing ) after argument list`
This happens because the quotes from the string interfere with the outer one. Maybe adding an escape is a better alternative?

choumx: Just add quotes?

jridgewell: Actually, I support using `JSON.stringify` here. It prevents the weird UA that includes a `'` or `"` char from breaking the string.

```js
JSON.stringify(`Chrome "test" 'ing'`);
// => `"Chrome \"test\" 'ing'"`

Issue #297, opened 2019-02-19
Multiple amp-script elements causes exception


fstanis:
```html
<amp-script layout=container src="/simple.js"><button id="hello">hello</button></amp-script>

<!-- Removing this prevents the error -->
<amp-script layout=container src="/empty.js"><button>hello</button></amp-script>
```

In `simple.js`:
```js
document.getElementById('hello').addEventListener('click', () => {});
```

On load, the following line throws a `Cannot read property 'onchange' of undefined` exception:

https://github.com/ampproject/worker-dom/blob/ec1decd8961372e0a571c2a3b1f2ee40d9f1587d/src/main-thread/commands/event-subscription.ts#L107

dvoytenko:
This is due to the fact that some data structures in https://github.com/ampproject/worker-dom/blob/master/src/main-thread/nodes.ts#L21 and elsewhere are module-scoped (e.g. `let NODES: Map<number, Node>` and `let BASE_ELEMENT: HTMLElement`). Runtime-wise they workout to be globally-scoped. Thus they are being overwritten by different instances of `amp-experiment` and thus there are many race conditions and other types of conflicts.

PR #299, opened 2019-02-21
Support multiple worker-dom roots in the main thread

dvoytenko:
Fixes: #297

Most of the main-thread code is now scoped per `baseElement` or `NodeContext`.

The one issue I haven't yet resolved - is treatment of `DebuggingContext` which currently bundled into main `install.ts` and adds about 0.6-0.7Kb.

/to @kristoferbaxter @choumx

### DIFF
--- a/src/main-thread/commands/bounding-client-rect.ts
+++ b/src/main-thread/commands/bounding-client-rect.ts
@@ -22,16 +22,16 @@ import { NumericBoolean } from '../../utils';
 import { WorkerContext } from '../worker';
 
 export class BoundingClientRectProcessor {
-  nodeContext_: NodeContext;
-  workerContext_: WorkerContext;
+  private nodeContext: NodeContext;
+  private workerContext: WorkerContext;
 
   /**
    * @param nodeContext
    * @param workerContext whom to dispatch events toward.
    */
   constructor(nodeContext: NodeContext, workerContext: WorkerContext) {
-    this.nodeContext_ = nodeContext;
-    this.workerContext_ = workerContext;
+    this.nodeContext = nodeContext;
+    this.workerContext = workerContext;
   }
 
   /**
@@ -40,7 +40,7 @@ export class BoundingClientRectProcessor {
    */
   process(mutation: TransferrableMutationRecord): void {
     const nodeId = mutation[TransferrableKeys.target];
-    const target = this.nodeContext_.getNode(nodeId);
+    const target = this.nodeContext.getNode(nodeId);
 
     if (!target) {
       console.error('getNode() yields a null value. Node id (' + nodeId + ') was not found.');
@@ -48,7 +48,7 @@ export class BoundingClientRectProcessor {
     }
 
     const boundingRect = target.getBoundingClientRect();
-    this.workerContext_.messageToWorker({
+    this.workerContext.messageToWorker({
       [TransferrableKeys.type]: MessageType.GET_BOUNDING_CLIENT_RECT,
       [TransferrableKeys.target]: {
         [TransferrableKeys.index]: target._index_,
--- a/src/main-thread/commands/bounding-client-rect.ts
+++ b/src/main-thread/commands/bounding-client-rect.ts
@@ -16,33 +16,52 @@
 
 import { TransferrableMutationRecord } from '../../transfer/TransferrableRecord';
 import { TransferrableKeys } from '../../transfer/TransferrableKeys';
-import { getNode } from '../nodes';
-import { messageToWorker } from '../worker';
 import { MessageType } from '../../transfer/Messages';
+import { NodeContext } from '../nodes';
 import { NumericBoolean } from '../../utils';
+import { WorkerContext } from '../worker';
 
-/**
- * Process commands transfered from worker thread to main thread.
- * @param nodesInstance nodes instance to execute commands against.
- * @param worker whom to dispatch events toward.
- * @param mutation mutation record containing commands to execute.
- */
-export function process(worker: Worker, mutation: TransferrableMutationRecord): void {
-  const nodeId = mutation[TransferrableKeys.target];
-  const target = getNode(nodeId);
+export class BoundingClientRectProcessor {
+  nodeContext_: NodeContext;
+  workerContext_: WorkerContext;
 
-  if (!target) {
-    console.error('getNode() yields a null value. Node id (' + nodeId + ') was not found.');
-    return;
+  /**
+   * @param nodeContext
+   * @param workerContext whom to dispatch events toward.
+   */
+  constructor(nodeContext: NodeContext, workerContext: WorkerContext) {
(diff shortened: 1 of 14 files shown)


src/main-thread/install.ts
```diff
@@ -30,41 +32,51 @@ const ALLOWABLE_MESSAGE_TYPES = [MessageType.MUTATE, MessageType.HYDRATE];
...
 export function install(
   fetchPromise: Promise<[string, string, string]>,
   baseElement: HTMLElement,
   callbacks?: WorkerCallbacks,
   sanitizer?: Sanitizer,
+  debug?: boolean,
 ): void {
-  prepareNodes(baseElement);
+  const strings = new Strings();
+  const nodeContext = new NodeContext(strings, baseElement);
```
choumx: One way to extract debugging is to force callers to provide/inject `Strings` and `NodeContext` (or just `Translator` if we combine them per suggestion above).

src/main-thread/nodes.ts
```diff
@@ -16,105 +16,117 @@
...
 import { TransferrableKeys } from '../transfer/TransferrableKeys';
-import { get as getString } from './strings';
+import { Strings } from './strings';
 
-let count: number = 2;
-let NODES: Map<number, Node>;
-let BASE_ELEMENT: HTMLElement;
+export class NodeContext {
+  baseElement_: HTMLElement;
+  strings_: Strings;
```
choumx: I think we can consider combining `Strings` and `NodeContext` since they both are basically lookup tables for data passed between page and worker. Seems like most classes require an instance of each anyways.

Suggested name: `Translator`?

src/main-thread/mutator.ts
```diff
@@ -18,33 +18,96 @@ import { TransferrableMutationRecord } from '../transfer/TransferrableRecord';
...
+import { Strings } from './strings';
+import { WorkerContext } from './worker';
+import { EventSubscriptionProcessor } from './commands/event-subscription';
+import { BoundingClientRectProcessor } from './commands/bounding-client-rect';
 
-let MUTATION_QUEUE: Array<TransferrableMutationRecord> = [];
-let PENDING_MUTATIONS: boolean = false;
-let worker: Worker;
-let sanitizer: Sanitizer | undefined;
+export class MutatorContext {
```
choumx: Minor: `MutationProcessor`?

src/main-thread/commands/bounding-client-rect.ts
```diff
@@ -16,33 +16,52 @@
...
- * @param nodesInstance nodes instance to execute commands against.
- * @param worker whom to dispatch events toward.
- * @param mutation mutation record containing commands to execute.
- */
-export function process(worker: Worker, mutation: TransferrableMutationRecord): void {
-  const nodeId = mutation[TransferrableKeys.target];
-  const target = getNode(nodeId);
+export class BoundingClientRectProcessor {
+  nodeContext_: NodeContext;
+  workerContext_: WorkerContext;
```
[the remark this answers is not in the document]
kristoferbaxter: +1, also no need to suffix with an underscore.

src/main-thread/commands/bounding-client-rect.ts
```diff
@@ -16,33 +16,52 @@
...
-/**
- * Process commands transfered from worker thread to main thread.
- * @param nodesInstance nodes instance to execute commands against.
- * @param worker whom to dispatch events toward.
- * @param mutation mutation record containing commands to execute.
- */
-export function process(worker: Worker, mutation: TransferrableMutationRecord): void {
-  const nodeId = mutation[TransferrableKeys.target];
-  const target = getNode(nodeId);
+export class BoundingClientRectProcessor {
```
kristoferbaxter: There might be a base class to extend here if the pattern repeats.

src/main-thread/mutator.ts
```diff
@@ -18,33 +18,96 @@ import { TransferrableMutationRecord } from '../transfer/TransferrableRecord';
...
+    }
+  }
+
+  /**
+   * Apply all stored mutations syncronously. This method works well, but can cause jank if there are too many
+   * mutations to apply in a single frame.
+   *
+   * Investigations in using asyncFlush to resolve are worth considering.
+   */
+  syncFlush_(): void {
```
kristoferbaxter: A good pattern to get into IMO is to always declare if a method is public or private in TS.

src/main-thread/commands/bounding-client-rect.ts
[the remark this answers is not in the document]
dvoytenko: I added `private` everywhere and removed suffixes. But I think we should use some for of a suffix/prefix. IMHO they still have value even with the explicit `private` keyword.

src/main-thread/commands/bounding-client-rect.ts
```diff
@@ -16,33 +16,52 @@
...
+  workerContext_: WorkerContext;
 
-  if (!target) {
-    console.error('getNode() yields a null value. Node id (' + nodeId + ') was not found.');
-    return;
+  /**
+   * @param nodeContext
+   * @param workerContext whom to dispatch events toward.
+   */
+  constructor(nodeContext: NodeContext, workerContext: WorkerContext) {
```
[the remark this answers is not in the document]
dvoytenko: Not yet sure if I like this style. But if you feel good about it, I'll do it. LMK.

src/main-thread/install.ts
dvoytenko: Sorry, what Translator and what's the suggestion above?

src/main-thread/mutator.ts
dvoytenko: Done

src/main-thread/commands/bounding-client-rect.ts
dvoytenko: Agree. Will wait a bit more to see how this develops though.

src/main-thread/mutator.ts
dvoytenko: Made it `private`. And agree. Though see above - still think we should have a prefix/suffix. But up to you.

src/main-thread/nodes.ts
dvoytenko: Ok. Understood. I actually thinking about a slightly different abstraction. Let me consider it a bit, if you don't mind. I'll probably push some more changes on this in a separate PR.

src/main-thread/commands/bounding-client-rect.ts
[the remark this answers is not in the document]
choumx: I think we can keep underscores. From `go/ts-sg`:

>  Identifiers must not use _ as a prefix or suffix. The exception is TypeScript code that is used by JavaScript programmers outside of google3, e.g. code published on npm.

choumx:
Does Dima have merge access yet? Happy to help merge too.

PR #308, opened 2019-02-26
Implement set innerHTML.

### LINKED ISSUE #283: Implement set innerHTML
Needed by #19, #167 and #282. 

Also used by some [React examples](https://reactjs.org/community/examples.html) I've tried e.g. [product-compare-react](https://github.com/Rhymond/product-compare-react).

We should also discuss how to handle the async problem. For measurement, we can mark elements that set `innerHTML` as "dirty" until reconciled from main thread, and output warnings when reading from "dirty" elements.

/to @delima02 /cc @kristoferbaxter @jridgewell

delima02:
- Adapt an HTML parser, written in TypeScript, to use our implementations of HTML elements. 
- Source: https://github.com/taoqf/node-html-parser
- Implement set innerHTML by parsing the text with the adapted version of the HTML parser and appending all extracted nodes to the desired root Element Node.
- Create unit tests for this behavior.
- Fixes #283

/cc @jridgewell @choumx


delima02:
This PR increases the bundle sizes of the following:
- `./dist/worker.mjs` by 0.63KB above bundle `maxSize`
- `./dist/worker.safe.mjs` by 0.1KB above bundle `maxSize`

config/rollup.worker-thread.js
```diff
@@ -208,4 +218,4 @@ const debugModules = DEBUG_BUNDLE_VALUE
     ]
   : [];
 
-export default [...ESModules, ...IIFEModules, ...debugModules];
+export default [...ESModules, /*...IIFEModules, ...debugModules*/];
```
jridgewell: Undo.

config/rollup.worker-thread.js
```diff
@@ -32,6 +34,7 @@ const ESModules = [
       sourcemap: true,
     },
     plugins: [
+      resolve(),
```
jridgewell: Is resolve actually necessary?

config/rollup.worker-thread.js
```diff
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+import resolve from 'rollup-plugin-node-resolve';
+import commonjs from 'rollup-plugin-commonjs';
```
jridgewell: Seems unused?

demo/hello-world/hello-world.js
```diff
@@ -23,4 +23,5 @@ btn.addEventListener('click', async () => {
 
   const boundingClientRect = await h1.getBoundingClientRectAsync();
   h1.textContent = h1.textContent + JSON.stringify(boundingClientRect);
+
```
jridgewell: Nit: Undo

src/test/element/innerHTML.ts
```diff
@@ -66,3 +67,67 @@ test('element with comment', t => {
   node.appendChild(comment);
   t.is(node.innerHTML, '<!--comment-->');
 });
+
+test('set nothing', t => {
```
jridgewell: Add a test that any children already present are removed

src/test/element/innerHTML.ts
```diff
@@ -66,3 +67,67 @@ test('element with comment', t => {
...
+});
+
+test('set an element node', t => {
+  const { node } = t.context;
+  node.innerHTML = '<div></div>';
+  t.is(node.childNodes.length, 1);
+
+  const child = node.firstChild;
+  if (child) {
+    t.is(child.nodeType, NodeType.ELEMENT_NODE);
```
jridgewell: Assert child is a div

src/test/element/innerHTML.ts
```diff
@@ -66,3 +67,67 @@ test('element with comment', t => {
...
+  t.is(node.hasChildNodes(), false);
+});
+
+test('set an element node', t => {
+  const { node } = t.context;
+  node.innerHTML = '<div></div>';
+  t.is(node.childNodes.length, 1);
+
+  const child = node.firstChild;
+  if (child) {
```
jridgewell: I would remove the if statement surrounding all of these.

src/test/element/innerHTML.ts
```diff
@@ -66,3 +67,67 @@ test('element with comment', t => {
...
+});
+
+test('set a text node', t => {
+  const { node } = t.context;
+  node.innerHTML = 'Hello World!';
+  t.is(node.childNodes.length, 1);
+
+  const child = node.firstChild;
+  if (child) {
+    t.is(child.nodeType, NodeType.TEXT_NODE);
```
jridgewell: Assert text content

src/test/element/innerHTML.ts
```diff
@@ -66,3 +67,67 @@ test('element with comment', t => {
...
+});
+
+test('set comment node', t => {
+  const { node } = t.context;
+  node.innerHTML = '<!--this is a comment-->';
+  t.is(node.childNodes.length, 1);
+
+  const child = node.firstChild;
+  if (child) {
+    t.is(child.nodeType, NodeType.COMMENT_NODE);
```
jridgewell: Assert comment contents

src/third_party/html-parser/LICENSE
```diff
@@ -0,0 +1,21 @@
+The MIT License (MIT)
```
jridgewell:
```suggestion
Copyright 2019 Xiaoyi Shi <ashi009@gmail.com>
```

src/third_party/html-parser/LICENSE
```diff
@@ -0,0 +1,21 @@
...
+
```
jridgewell: Additionally, we need a `README.md` file like https://github.com/ampproject/amphtml/blob/master/third_party/css-escape/README.amp

src/third_party/html-parser/LICENSE
```diff
@@ -0,0 +1,21 @@
...
+THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND,
+EXPRESS OR IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF
+MERCHANTABILITY, FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO
+EVENT SHALL THE AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR
+OTHER LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING
+FROM, OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+THE SOFTWARE.
+
+https://opensource.org/licenses/MIT
```
jridgewell: Delete this line

src/test/element/innerHTML.ts
delima02: I added these because otherwise it would throw, specifying how (child) `Object is possibly 'null'`

src/test/element/innerHTML.ts
delima02: Good catch :) actually missed this behavior.

config/rollup.worker-thread.js
delima02: Was at some point, now it doesn't seem to be. I think it was needed with the `he` import we had.

demo/hello-world/hello-world.js
```diff
@@ -20,7 +20,7 @@ btn.addEventListener('click', async () => {
   const h1 = document.createElement('h1');
   h1.textContent = 'Hello World!'
   document.body.appendChild(h1);
-
   const boundingClientRect = await h1.getBoundingClientRectAsync();
   h1.textContent = h1.textContent + JSON.stringify(boundingClientRect);
+
```
jridgewell: Nit: Undo changes to this file.

src/test/element/innerHTML.ts
jridgewell: Use typescript's [non-null assertion](https://github.com/Microsoft/TypeScript/wiki/What's-new-in-TypeScript#non-null-assertion-operator): `const child = node.firstChild!;`

src/worker-thread/dom/Element.ts
```diff
@@ -169,6 +170,24 @@ export class Element extends ParentNode {
...
 
+  /**
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML
+   * @param html The raw html string to parse.
+   */
+  set innerHTML(html: string) {
+    const root = parse(html, this);
+
+    // remove previous children
+    this.childNodes.forEach(n => {
```
jridgewell: This'll cause an n<sup>2</sup> removal: https://github.com/ampproject/worker-dom/blob/0f3eeed0ecec3e4a6ec7c9de53073283902e8422/src/worker-thread/dom/Node.ts#L267

src/third_party/html-parser/LICENSE
jridgewell: Ping.

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,221 @@
...
+        } else {
+          lastTextPos = kMarkupPattern.lastIndex = index + closeMarkup.length;
+          match[1] = 'true';
+        }
+      }
+    }
+    if (match[1] || match[4] || kSelfClosingElements[match[2]]) {
+      // </ or /> or <br> etc.
+      while (true) {
+        if (currentParent.nodeName == match[2]) {
```
jridgewell: This is due to `kSelfClosingElements[match[2]]`. It seems like it would be better to separate that out of the loop entirely.

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,221 @@
...
+            }
+          }
+          // Use aggressive strategy to handle unmatching markups.
+          break;
+        }
+      }
+    }
+  }
+  type Response = (Element | Text) & { valid: boolean };
+  const valid = !!(stack.length === 1);
```
jridgewell: Nit: Remove the `!!`

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,221 @@
...
+
+  if (!valid) {
+    throw new Error('Attempting to parse invalid tag.');
+  }
+
+  const response = root as Response;
+  response.valid = valid;
```
jridgewell: Let's return a struct instead of adding a new prop to the `Element`.

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,221 @@
...
+  while (stack.length > 1) {
```
jridgewell: Drop this while statement and everything in it, we don't need it.

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,221 @@
...
+    }
+  }
+  response.childNodes.forEach(node => {
+    if (node instanceof Element) {
+      node.parentNode = null;
+    }
+  });
+
+  // remove the added <div>
+  if (response && response.firstChild) {
```
jridgewell: This should be unnecessary after the `lastTextPos` change.

src/worker-thread/dom/Element.ts
delima02: Is the new approach better?

src/third_party/html-parser/html-parser.ts
delima02: I attempted doing this change and it does not work (it follows the same behavior as if the default value was `-1` and the wrapping `<div>` wasn't added

src/third_party/html-parser/html-parser.ts
delima02: Not sure if this is what you meant

jridgewell: We already have `Element`, and its instances have known properties. We're adding a foreign property to it. We shouldn't be doing that.

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,221 @@
...
+  [key: string]: string;
+}
+
+function arr_back<T>(arr: T[]) {
+  return arr[arr.length - 1];
+}
+
+// https://html.spec.whatwg.org/multipage/custom-elements.html#valid-custom-element-name
+const kMarkupPattern = /<!--[^]*?(?=-->)-->|<(\/?)([a-z][-.0-9_a-z]*)\s*([^>]*?)(\/?)>/gi;
+const kAttributePattern = /(^|\s)(id|class)\s*=\s*("([^"]+)"|'([^']+)'|(\S+))/gi;
```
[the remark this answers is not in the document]
delima02: Gave my best shot looking at doing this by looking at 
[https://html.spec.whatwg.org/multipage/syntax.html#attributes-2](url).
It seems that we don't accept certain attribute names that according to the source mentioned should be. We should discuss this.

src/third_party/html-parser/html-parser.ts
delima02: Used a JS object as response. Does that sound like OK?

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,221 @@
...
+
+      stack.push(currentParent);
+      if (kBlockTextElements[match[2]]) {
+        // a little test to find next </script> or </style> ...
+        let closeMarkup = '</' + match[2] + '>';
+        let index = data.indexOf(closeMarkup, kMarkupPattern.lastIndex);
+        if (index == -1) {
+          lastTextPos = kMarkupPattern.lastIndex = data.length + 1;
+        } else {
+          lastTextPos = kMarkupPattern.lastIndex = index + closeMarkup.length;
```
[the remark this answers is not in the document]
delima02: Not sure I follow here...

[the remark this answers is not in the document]
delima02: Okay I see your point. The question here is, do we actually want block text elements? We should discuss the desired behavior for these tags (especially `<script>`).

src/third_party/html-parser/html-parser.ts
```diff
@@ -115,21 +104,19 @@ export function parse(data: string, rootElement: Element) {
...
         }
       }
-      const childToAppend = new Element(NodeType.ELEMENT_NODE, match[2], currentParent.namespaceURI, currentParent.ownerDocument);
+      const childToAppend = new Element(currentParent.nodeType, match[2], currentParent.namespaceURI, ownerDocument);
 
-      for (const key in attrs) {
-        childToAppend.setAttribute(key, attrs[key]);
+      let attrs: Attributes = {};
+      for (let attMatch; (attMatch = kAttributePattern.exec(match[3])); ) {
+        attrs[attMatch[2]] = attMatch[4] || attMatch[5] || attMatch[6];
```
jridgewell: This is useless now.

src/third_party/html-parser/html-parser.ts
```diff
@@ -141,6 +128,7 @@ export function parse(data: string, rootElement: Element) {
         let index = data.indexOf(closeMarkup, kMarkupPattern.lastIndex);
         if (index == -1) {
           lastTextPos = kMarkupPattern.lastIndex = data.length + 1;
```
jridgewell: No need for this line if we're gonna throw and error. 😉

src/third_party/html-parser/html-parser.ts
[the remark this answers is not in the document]
jridgewell: `style` and `pre` definitely need this. We can discuss `script` in https://github.com/ampproject/worker-dom/issues/311.

src/worker-thread/dom/Element.ts
jridgewell: Yup. But is there a way to do this without the copy paste?

src/worker-thread/dom/Element.ts
```diff
@@ -175,13 +176,28 @@ export class Element extends ParentNode {
    * @param html The raw html string to parse.
    */
   set innerHTML(html: string) {
-    const root = parse(html, this);
+    const parsedResponse = parse(html, this);
+
+    if (!parsedResponse.valid) {
+      throw new Error('Attempting to set invalid HTML content.');
```
jridgewell: This should be done in the parser.

src/third_party/html-parser/html-parser.ts
```diff
@@ -169,52 +157,24 @@ export function parse(data: string, rootElement: Element) {
...
+  root.childNodes.forEach(node => {
     if (node instanceof Element) {
       node.parentNode = null;
     }
   });
 
   // remove the added <div>
-  if (response && response.firstChild) {
-    response.firstChild.childNodes.forEach(node => {
+  if (root.firstChild) {
```
jridgewell: I still think we'd be able to remove this if we removed https://github.com/ampproject/worker-dom/pull/308/files/38bfb90146410e95c1267fcaa7c5c162c7a91971..f2a2335c905b54493f635634e18d36af035cef93#diff-61c6395322f1f65ada0d1a9c981948ffR84

src/worker-thread/dom/Element.ts
delima02: Thought about adding a method in Node class that would remove all. Since it's not part of Node ([https://developer.mozilla.org/en-US/docs/Web/API/Node](url)), thought it'd be better not to do it.

src/worker-thread/dom/Element.ts
delima02: Being that the case do we even want to include `valid` in the response then?

src/worker-thread/dom/Element.ts
jridgewell: Eh, don't worry about it.

src/worker-thread/dom/Element.ts
jridgewell: Probably not.

src/third_party/html-parser/html-parser.ts
delima02: This conditional actually serves another purpose (to match closing tags for `currentParent`). I can remove `kSelfClosingElements[tagName]` from the outer `if` statement, but I can't remove this conditional (there would be unmatched tags and algorithm wouldn't work). Should I still do it?

src/test/element/innerHTML.ts
```diff
@@ -66,3 +67,111 @@ test('element with comment', t => {
...
+test('set keeps tagName\'s case', t => {
+  const { node } = t.context;
+  node.innerHTML = '<svg><feImage></feImage></svg>';
+  const svgWrapper = node.firstChild!;
+  const child = svgWrapper.firstChild!;
+  t.is(child.nodeName, 'feImage');
+});
+
+test.skip('set has svg tags live in SVG namespace', t => {
+  t.pass();
```
jridgewell: Delete

src/test/element/innerHTML.ts
```diff
@@ -66,3 +67,111 @@ test('element with comment', t => {
...
+  const { node } = t.context;
+  node.innerHTML = '<svg></svg>';
+  const child = node.firstChild!;
+  t.is(child.namespaceURI, SVG_NAMESPACE);
+});
+
+test.skip('set normalizes html namespace tag names', t => {
+  t.pass();
```
jridgewell: Ditto

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,190 @@
...
+  const stack = [root as Node];
+  let lastTextPos = 0;
+  let match: RegExpExecArray | null;
+  data = '<div>' + data + '</div>';
+
+  while ((match = kMarkupPattern.exec(data))) {
+
+    const commentContents = match[1]; // <!--contents-->
+    let beginningSlash = match[2]; // ... </ ...
+    let tagName = match[3];
```
jridgewell: const

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,190 @@
...
+    const matchAttributes = match[4];
+    const endSlash = match[5]; // ... /> ...
+
+    if (lastTextPos < match.index) {
+      // if has content
+      const text = data.slice(lastTextPos, match.index);
+      currentParent.appendChild(new Text(text, ownerDocument));
+    }
+    lastTextPos = kMarkupPattern.lastIndex;
+    if (match[0][1] == '!') {
```
jridgewell: This should check for `commentContents !== undefined`

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,190 @@
...
+      currentParent.appendChild(new Comment(commentContents, ownerDocument));
+      continue;
+    }
+
+    const normalizedTagName = tagName.toUpperCase();
+
+    if (!beginningSlash) {
+      // not </ tags
+
+      if (!endSlash && kElementsClosedByOpening[currentParent.normalizedTagName]) {
```
jridgewell: `currentParent` doesn't have a `normalizedTagName`. Tests for this are needed.

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,190 @@
...
+        const attrValue = attMatch[4] || attMatch[5] || attMatch[6];
+        childToAppend.setAttribute(attrName, attrValue);
+      }
+
+      currentParent = currentParent.appendChild(childToAppend);
+
+      stack.push(currentParent);
+      if (kBlockTextElements[normalizedTagName]) {
+        // a little test to find next </script> or </style> ...
+        let closeMarkup = '</' + normalizedTagName.toLowerCase() + '>';
```
jridgewell: const

src/third_party/html-parser/html-parser.ts
```diff
@@ -0,0 +1,190 @@
...
+        let index = data.indexOf(closeMarkup, kMarkupPattern.lastIndex);
+        if (index == -1) {
+          throw new Error('Close markup not found.');
+        } else {
+          const text = data.slice(lastTextPos, index);
+          currentParent.appendChild(new Text(text, ownerDocument));
```
jridgewell: We already have a code path for this above. We should adjust the `kMarkupPattern.lastIndex` so it will match for our closing tag, and let the normal loop.

src/test/element/innerHTML.ts
```diff
@@ -159,6 +159,20 @@ test('set keeps tagName\'s case', t => {
   t.is(child.nodeName, 'feImage');
 });
 
+test('set closes tags by opening others', t => {
+  const { node } = t.context;
+  node.innerHTML = '<b><div></div>';
+  // <b> tag should be closed to create a second child
```
jridgewell: Hmm, this seems incorrect in the real DOM. It seems the `<div>` becomes a child of the `<b>`.

src/third_party/html-parser/html-parser.ts
```diff
@@ -101,11 +102,9 @@ export function parse(data: string, rootElement: Element) {
 
     if (!beginningSlash) {
       // not </ tags
-
       if (!endSlash && kElementsClosedByOpening[currentParent.tagName]) {
         if (kElementsClosedByOpening[currentParent.tagName][normalizedTagName]) {
-          stack.pop();
-          currentParent = arr_back(stack);
+          tagsClosed.push(currentParent.tagName);
```
jridgewell: I don't believe this is the correct solution. The issue here is that opening a `<b>` does not imply that any other tag is closed. See https://www.w3.org/TR/html5/syntax.html#optional-tags for a full lit of implied-end-by-opening and implied-end-by-closing.

src/test/element/innerHTML.ts
```diff
@@ -66,3 +67,130 @@ test('element with comment', t => {
...
+  const child = node.firstChild!;
+  t.is(child.namespaceURI, SVG_NAMESPACE);
+});
+
+test.skip('set keeps tagName\'s case', t => {
+  const { node } = t.context;
+  node.innerHTML = '<svg><feImage></feImage></svg>';
+  const svgWrapper = node.firstChild!;
+  const child = svgWrapper.firstChild!;
+  t.is(child.nodeName, 'feImage');
```
choumx: Why skip these tests?

src/test/element/innerHTML.ts
```diff
@@ -66,3 +67,130 @@ test('element with comment', t => {
...
+  t.is(childContent.nodeType, NodeType.TEXT_NODE);
+});
+
+test('set normalizes html namespace tag names', t => {
+  const { node } = t.context;
+  node.innerHTML = '<Div></Div>';
+  const child = node.firstChild!;
+  t.is(child.localName, 'div');
+  t.is(child.nodeName, 'DIV');
+});
```
choumx: Any reason why we're not just testing against get `innerHTML` in all of these tests? I.e.

```ts
t.is(node.innerHTML, '<div></div>');
```

choumx:
Merging to get this in first. @delima02 a couple questions above for you.

choumx:
Looks like circular deps are still an issue?

> (!) Circular dependency: output/worker-thread/dom/Element.js -> output/third_party/html-parser/html-parser.js -> output/worker-thread/dom/Element.js

src/test/element/innerHTML.ts
delima02: I was thinking, what if something goes on with `get innerHTML`? That would make all tests for `set innerHTML` fail as well if we did this. I'm just trying to be as modular as possible.

src/test/element/innerHTML.ts
delima02: @jridgewell

src/test/element/innerHTML.ts
choumx: Eh, modularizing a getter vs. the setter of the same function doesn't buy us much. The test code would be much more readable the other way.

PR #318, opened 2019-03-01
Completely separate main-thread from worker-thread code

dvoytenko:


src/main-thread/install.ts
```diff
@@ -75,7 +74,6 @@ export function install(
     if (workerDOMScript && authorScript && authorScriptURL) {
       const workerContext = new WorkerContext(baseElement, workerDOMScript, authorScript, authorScriptURL, callbacks);
       const worker = workerContext.getWorker();
-      setPhase(Phases.Hydrating);
```
choumx: Where does this happen now? Or is the worker phase initialized to `Hydrating`?

dvoytenko: Only worker ever uses it this way. And that code is not isolated. So I'd rather not to have it here at all.

PR #320, opened 2019-03-01
Mutation pump

dvoytenko:
The use of this API is roughly like this:

```
install(..., {
  onMutationPump: (callback, phase) => {
    if (phase == HYDRATING) {
      // Always allow hydrations.
      callback();
    } else if (shouldAllowMutation()) {
      // Otherwise, more selective pumping.
      callback();
    }
  },
});
```

Thus, this model can:

1. Decide whether allow mutations or not.
2. Prioritize mutations differently. E.g. it can slow down the mutation pump when an element is backgrounded, etc.


src/main-thread/mutator.ts
```diff
@@ -62,23 +78,23 @@ export class MutatorProcessor {
...
+  ): void {
+    this.stringQueue = this.stringQueue.concat(stringValues);
+    this.nodeQueue = this.nodeQueue.concat(nodes);
     this.mutationQueue = this.mutationQueue.concat(mutations);
-    if (!this.pendingMutations) {
-      this.pendingMutations = true;
-      requestAnimationFrame(this.syncFlush.bind(this));
+    if (!this.pendingQueue) {
+      this.pendingQueue = true;
+      this.mutationPump(this.boundSyncFlush, phase);
```
choumx: Drop-in compatibility with `requestAnimationFrame` is nice but I think it'd be cleaner to avoid "callback-in-callback" and just use a return value.

```js
install(..., {
  onMutationPump: (phase) => {
    if (phase == HYDRATING) {
      // Always allow hydrations.
      return true;
    } else {
      // Otherwise, more selective pumping.
      return shouldAllowMutation();
    }
  },
});
```

dvoytenko: The reason I picked that option is because I wanted this method to also support an alternative queuing. Do you think it won't be needed? Some ideas we had assumed a possibility of re-queueing until the next action.

choumx: Either this or a separate flush trigger SGTM. 

TIL "[message pump](https://en.wikipedia.org/wiki/Event_loop#Message_passing)" ~= "event loop". :)

dvoytenko: @choumx Added more documentation. Let me know if this makes sense to you.

src/transfer/TransferrablePhase.ts
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+export const enum TransferrablePhase {
+  Initializing = 0,
+  Hydrating = 1,
+  Mutating = 2,
+}
```
choumx: Minor: Possible to just move src/worker-thread/phases.ts here and avoid two identical enums? "TransferrablePhase" makes me think this is data that will be transferred across `postMessage`.

src/main-thread/callbacks.ts
```diff
@@ -15,6 +15,14 @@
...
 
 import { MessageFromWorker, MessageToWorker } from '../transfer/Messages';
+import { TransferrablePhase } from '../transfer/TransferrablePhase';
+
+/**
+ * The callback for `onMutationPump`. If specified, this callback will be called
+ * for the new set of mutations pending. The callback can either immediately
+ * call `flush()`, or it can reject mutations, or it can batch them further.
+ */
+export type MutationPumpFunction = (flush: Function, phase: TransferrablePhase) => void;
```
choumx: +1 to "flush".

src/main-thread/mutator.ts
```diff
@@ -40,12 +46,22 @@ export class MutatorProcessor {
...
     this.strings = strings;
     this.nodeContext = nodeContext;
     this.sanitizer = passedSanitizer;
+    this.mutationPump = mutationPump || requestAnimationFrame.bind(null);
+    this.boundSyncFlush = this.syncFlush.bind(this);
+    this.stringQueue = [];
+    this.nodeQueue = [];
     this.mutationQueue = [];
-    this.pendingMutations = false;
+    this.pendingQueue = false;
```
choumx: Can we derive `pendingQueue` from the length of the queue arrays?

src/main-thread/mutator.ts
```diff
@@ -40,12 +46,22 @@ export class MutatorProcessor {
...
+    mutationPump?: MutationPumpFunction,
+    passedSanitizer?: Sanitizer,
+  ) {
     this.strings = strings;
     this.nodeContext = nodeContext;
     this.sanitizer = passedSanitizer;
+    this.mutationPump = mutationPump || requestAnimationFrame.bind(null);
+    this.boundSyncFlush = this.syncFlush.bind(this);
+    this.stringQueue = [];
+    this.nodeQueue = [];
```
choumx: Combine into single `queue` of `Array<{mutation: Array, strings: Array, nodes: Array}>`?

dvoytenko: Thought about this. The main negative is the new object allocation for each push operation. Not sure what's the nominal frequency here is. Please advise.

src/transfer/TransferrablePhase.ts
dvoytenko: Done

src/main-thread/mutator.ts
dvoytenko: It'd be somewhat harder to read, imho, b/c i'll have to be precalculating this value from queue length before processing mutation. Current code pushes queue, then checks if it was scheduled, and then schedules.

src/main-thread/mutator.ts
choumx: Looks like the worker's `MutationObserver` batches mutations by Promise tick so I don't imagine the allocation being an issue. 

Mostly, I'm wary of bugs from out-of-sync variables. Up to you though.

package.json
```diff
@@ -50,6 +51,7 @@
     "ava": "1.2.1",
     "babel-plugin-minify-replace": "0.5.0",
     "babel-plugin-transform-remove-console": "6.9.4",
+    "browser-env": "^3.2.6",
```
choumx: Looks like the creators of ava/jsdom recommend [window](https://github.com/lukechilds/window) instead of browser-env.

Context: https://github.com/avajs/ava/issues/1092

src/test/main-thread/tsconfig.json
```diff
@@ -0,0 +1,23 @@
...
+    "strictNullChecks": true,
+    "noImplicitAny": true,
+    "noImplicitThis": true,
+    "alwaysStrict": true
+  },
+  "include": [
+    "**/*.ts",
+    "../../main-thread/*.d.ts"
+  ]
+}
```
choumx: Can this tsconfig extend `src/test/tsconfig.json` and just overwrite `compilerOptions.lib`?

src/main-thread/mutator.ts
dvoytenko: I think out-of-sync bugs are well encapsulated by this class so I don't think it'd be very likely. But array-of-a-structure-of-arrays seems rather clumsy to me. It'd make the processing code less readable as well. So I'd rather keep it as three arrays. I'd consider redoing it if we'll need a solution where we need to apply a subset of a queue. But I don't think this is currently likely.

package.json
dvoytenko: Ok, this makes sense. I redid via jsdom. Happy to get rid of browser-env - seemed pretty heavy in terms of test performance.

src/test/main-thread/tsconfig.json
dvoytenko: Unfortunately that's not very convenient. I think a better option would be to organize all of `test` the same way as the rest of sources, e.g. `src/worker-thread` and `src/main-thread` - they really do have very little in common.

Issue #319, opened 2019-03-01
Preact browser test suite compatibility


choumx:
To build confidence in framework compat, run Preact's [browser test suite](https://github.com/developit/preact/tree/master/test/browser) against worker-dom.

Initial results (preact @ [3aaa578](https://github.com/developit/preact/commit/3aaa578), Feb 13): 
>  95 passing (13s)
  2 pending
  53 failing

https://gist.github.com/choumx/d44d5291ddddbcbe824349abceacf319

Many of these might be test config or test code issues, but some useful errors so far e.g. `document.createEvent` missing, nodeName returns "div" instead of "DIV", and no set `innerHTML`.

choumx:
With referenced PRs so far:

> 125 passing (13s)
2 pending
23 failing

choumx:
> 133 passing (12s)
2 pending
15 failing

choumx:
With above PRs and `set innerHTML` from #308:

> 145 passing (283ms)
2 pending
3 failing

choumx:
Pulled latest and ran again. Lots of changes in Preact recently with v10 branch being merged. Going to put a pin in this for now as we're in a reasonable state. 

I have 1-2 more upcoming fixes. Then, as of [0c8984b](https://github.com/developit/preact/commit/0c8984b) March 11:

> 286 passing (601ms)
20 pending
20 failing

Important tests like "Components", "render()" and "Lifecycle methods" are all green. 20 test failures include sections: "focus", "hydrate()", "keys", and "svg". [Full results here](https://gist.github.com/choumx/9d91834ce633ac1dd2f3dc969d67cc7c).

Issue #317, opened 2019-03-01
Type specificity in API parameters


choumx:
We may be too optimistic in ignoring JS vs. TS typing issues in APIs, e.g.

```ts
// Element.ts
public setAttribute(name: string, value: string): void {...}
```

E.g. there's no type conversion if a caller passes in a number or other non-string type.

```js
const el = document.createElement('p');
el.setAttribute('data-foo', 123);
el.getAttribute('data-foo'); // Expected: "123", actual: 123.
```

PR #321, opened 2019-03-01
Convert attribute values to strings

choumx:
Partial for #317. Fixes 6 test failures in #319.

### DIFF
--- a/src/worker-thread/dom/Element.ts
+++ b/src/worker-thread/dom/Element.ts
@@ -202,7 +202,7 @@ export class Element extends ParentNode {
    * @param name attribute name
    * @param value attribute value
    */
-  public setAttribute(name: string, value: string): void {
+  public setAttribute(name: string, value: Object | null | undefined): void {
     this.setAttributeNS(HTML_NAMESPACE, name, value);
   }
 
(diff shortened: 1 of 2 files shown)


src/worker-thread/dom/Element.ts
```diff
@@ -202,7 +202,7 @@ export class Element extends ParentNode {
    * @param name attribute name
    * @param value attribute value
    */
-  public setAttribute(name: string, value: string): void {
+  public setAttribute(name: string, value: Object | null | undefined): void {
```
jridgewell:
```suggestion
  public setAttribute(name: string, value: any): void {
```

src/worker-thread/dom/Element.ts
```diff
@@ -254,26 +254,27 @@ export class Element extends ParentNode {
    * @param name attribute name
    * @param value attribute value
    */
-  public setAttributeNS(namespaceURI: NamespaceURI, name: string, value: string): void {
+  public setAttributeNS(namespaceURI: NamespaceURI, name: string, value: Object | null | undefined): void {
```
jridgewell:
```suggestion
  public setAttributeNS(namespaceURI: NamespaceURI, name: string, value: any): void {
```

src/worker-thread/dom/Element.ts
jridgewell: Or, `unknown` if we lint against `any`.

choumx: `any` opts out of type checking, so we want `unknown` (TIL).

PR #322, opened 2019-03-01
tagName should be uppercase

choumx:
Fixes 2 test failures in #319. Also puts `tagName` into spec compliance:

> The tagName attribute’s getter must return the context object’s HTML-uppercased qualified name.

https://dom.spec.whatwg.org/#dom-element-tagname

Also added a couple `console.assert` which are effectively dev-only assertions.

PR #329, opened 2019-03-05
tagName should be uppercase (remix)

choumx:
Rolls forward #322 with fixes that required rollback in #327. Partial for #319.

- `document.createElement()` should always lowercases its input.
- `document.createElementNS()` should preserve casing of its input.
- `document.createElement('svg')` should not create `SVGElement`.
- Use `localName` instead of `tagName` for `document.createElement()` on main thread.

Tested on all workerDOM demos.


src/test/document/createElement.ts
```diff
@@ -0,0 +1,67 @@
...
+const test = anyTest as TestInterface<{
+  document: Document;
+}>;
+
+test.beforeEach(t => {
+  const document = createDocument();
+  t.context = { document };
+});
+
+test('createElement() should lowercase its input', t => {
```
delima02: Maybe assert that `createElement` will only create elements in HTML namespace?

src/worker-thread/initialize.ts
```diff
@@ -43,10 +43,8 @@ export function create(document: Document, strings: Array<string>, skeleton: Hyd
     default:
       const namespace: string | undefined =
         skeleton[TransferrableKeys.namespaceURI] !== undefined ? strings[skeleton[TransferrableKeys.namespaceURI] as number] : undefined;
-      const nodeName = strings[skeleton[TransferrableKeys.nodeName]];
-      const node: HTMLElement | SVGElement = namespace
```
jridgewell: Do we only transfer namespaceURI with svg elements?

choumx: We also transfer it for non-svg elements, which is why we got broken div elements e.g.

```js
document.createElement('DIV') // OK
document.createElementNS('http://www.w3.org/1999/xhtml', 'DIV') // NOPE
```

PR #330, opened 2019-03-05
Add details on querySelector(All) support

nainar:


web_compat_table.md
```diff
@@ -977,3 +977,5 @@ This section highlights the DOM APIs that are implemented in WorkerDOM currently
 | Window.visualViewport                               | ✖️     |                                                  | 
 | Window.window                                       | ✖️     |                                                  | 
 | Worker                                              | N/A    | Doesn't apply since this is on the Worker thread | 
+
+[^footnote]: querySelector(All) currently supports simple selectors i.e. ID selectors, Class selectors, Element selectors and Attribute selectors.
```
choumx: Is `[^footnote]` supported in GitHub flavored Markdown? I can't see it when clicking "View file".

nainar: Turns out you are right, changed to hyperlinks as per this stackoverflow thread: https://stackoverflow.com/questions/25579868/how-to-add-footnotes-to-github-flavoured-markdown

PR #332, opened 2019-03-05
Avoid super.getter

### LINKED ISSUE #326: textContent getter throws exception
It seems that calling the `textContent` getter throws `Uncaught TypeError: Cannot read property 'length' of undefined`, example code:

HTML:
```
<amp-script layout=container src="script.js">
  <button id="button">Run</button>
  <div id="input">Hello</div>
</amp-script>
```

JS:
```js
document.getElementById('button').addEventListener('click', () => {
  document.getElementById('input').textContent;
});
```

Exception seems to be thrown here:
https://github.com/ampproject/worker-dom/blob/f020b73f10f04d25b67c9d406386f3e14e337292/src/worker-thread/dom/Node.ts#L112

choumx:
Fixes #326. Also fixes bunch of tests for #319 due to Preact browser tests' use of `babel-register`.

#328 fixes the same issue for worker-dom's ES5 output, but not for projects that import the ES6 output and have their own transpilation pipeline (e.g. amphtml). Plus, the spec-compliant Babel transpile is pretty large.

### DIFF
--- a/src/worker-thread/dom/Element.ts
+++ b/src/worker-thread/dom/Element.ts
@@ -185,7 +185,7 @@ export class Element extends ParentNode {
    * @return text from all childNodes.
    */
   get textContent(): string {
-    return super.textContent;
+    return this.getTextContent();
   }
 
   /**
--- a/src/worker-thread/dom/Node.ts
+++ b/src/worker-thread/dom/Node.ts
@@ -106,6 +106,13 @@ export abstract class Node {
    * @return text from all childNodes.
    */
   get textContent(): string {
+    return this.getTextContent();
+  }
+
+  /**
+   * Use `this.getTextContent()` instead of `super.textContent` to avoid incorrect or expensive ES5 transpilation.
+   */
+  protected getTextContent(): string {
     let textContent = '';
     const childNodes = this.childNodes;
 


src/worker-thread/dom/Element.ts
```diff
@@ -185,7 +185,7 @@ export class Element extends ParentNode {
    * @return text from all childNodes.
    */
   get textContent(): string {
-    return super.textContent;
```
jridgewell: How many other occurrences are there? Seems like we'd need a linter to truly fix this.

choumx: None, but yea automation would be nice.

jridgewell: https://astexplorer.net/#/gist/39f577ac5181905df3258b719266ab1a/4430476dcea9304cc467bab9292c1a8091c21b79

choumx:
> would we be able to selectively apply this change only for the small population using transpiled to ES5 code?

Not sure how we can do that without compile-time flags e.g. `#ifdef` or similar.

jridgewell:
If we follow this `getX` convention: https://astexplorer.net/#/gist/1a29fc494a1d985e320004d31df72870/5c59bae1e25b8439b6674dfcbd5a159bff8f24a6

kristoferbaxter:
Justin's approach is what I was suggesting.

src/worker-thread/dom/Element.ts
choumx: #349

choumx:
Good ideas, but do we need to block this PR? Size impact is ~50 bytes. 

Filed #349 for the lint rule and rollup plugin.

kristoferbaxter:
I'm fine with moving along until we can address #349. Do you want to move back to loose mode as well for the `nomodule` output?

choumx:
Either works for me. I think loose mode may generate a smaller `nomodule` output.

PR #339, opened 2019-03-07
Fix Node.replaceChild when newChild is already a child

choumx:
Partial for #319.

Importantly, fixes this bug:
```js
parent.appendChild(one);
parent.appendChild(two);
parent.replaceChild(two, one);

// Expected: [one]
// Actual: [one, one]
log(parent.childNodes); 
```

Also fixes one case of [HierarchyRequestError](https://dom.spec.whatwg.org/#concept-node-replace) where newChild is an ancestor of parent.

### DIFF
--- a/src/test/node/replaceChild.ts
+++ b/src/test/node/replaceChild.ts
@@ -130,7 +130,7 @@ test('replacing a child with an ancestor', t => {
   t.deepEqual(parent.childNodes, [child]);
   t.deepEqual(child.childNodes, [grandchild]);
 
-  // Replacing with child should be a no-op.
+  // Replacing with ancestor should be a no-op.
   replaced = child.replaceChild(parent, grandchild);
   t.is(replaced, grandchild);
   t.is(child.parentNode, parent);
--- a/src/test/node/replaceChild.ts
+++ b/src/test/node/replaceChild.ts
@@ -80,7 +80,7 @@ test('replacing a child with another when there are multiple children', t => {
   t.deepEqual(parent.childNodes, [x, z]);
 });
 
-test('replacing a child with another child', t => {
+test('replacing a child with next sibling', t => {
   const { parent, x, y } = t.context;
 
   parent.appendChild(x);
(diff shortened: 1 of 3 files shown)


src/worker-thread/dom/Node.ts
```diff
@@ -277,35 +277,46 @@ export abstract class Node {
...
-        });
-      }
+    if (newChild === oldChild) {
+      return oldChild;
+    }
+    if (oldChild.parentNode !== this) {
+      // In DOM, this throws DOMException: "The node to be replaced is not a child of this node."
+      return oldChild;
+    }
+    if (newChild.contains(this)) {
```
kristoferbaxter: Should we combine these conditions then given they return the same result?

src/test/node/replaceChild.ts
```diff
@@ -79,3 +79,46 @@ test('replacing a child with another when there are multiple children', t => {
...
+
+test('replacing a child with another child', t => {
+  const { parent, x, y } = t.context;
+
+  parent.appendChild(x);
+  parent.appendChild(y);
+  t.is(x.parentNode, parent);
+  t.is(y.parentNode, parent);
+
+  const replaced = parent.replaceChild(y, x);
```
jridgewell: Add another test for doing this in the other order.

src/test/node/replaceChild.ts
```diff
@@ -79,3 +79,46 @@ test('replacing a child with another when there are multiple children', t => {
...
+
+  parent.appendChild(child);
+  child.appendChild(grandchild);
+  t.is(child.parentNode, parent);
+  t.is(grandchild.parentNode, child);
+  t.deepEqual(parent.childNodes, [child]);
+  t.deepEqual(child.childNodes, [grandchild]);
+
+  // Replacing with self should be a no-op.
+  let replaced = child.replaceChild(child, grandchild);
```
jridgewell: Shouldn't this throw?

src/test/node/replaceChild.ts
```diff
@@ -79,3 +79,46 @@ test('replacing a child with another when there are multiple children', t => {
...
+  t.is(replaced, grandchild);
+  t.is(child.parentNode, parent);
+  t.is(grandchild.parentNode, child);
+  t.deepEqual(parent.childNodes, [child]);
+  t.deepEqual(child.childNodes, [grandchild]);
+
+  // Replacing with child should be a no-op.
```
jridgewell: Bad comment

src/test/node/replaceChild.ts
kristoferbaxter: Right now `WorkerDOM` doesn't throw on DOM Exceptions. If you'd like to introduce this behaviour, let's make an issue to discuss.

PR #354, opened 2019-03-12
Install returns WorkerDom that allows termination

dvoytenko:


src/main-thread/worker.ts
```diff
@@ -39,7 +39,7 @@ export class WorkerContext {
     // TODO(KB): Minify this output during build process.
     const keys: Array<string> = [];
     const { skeleton, strings } = createHydrateableRootNode(baseElement);
-    for (const key in document.body.style) {
+    for (const key in baseElement.style) {
```
kristoferbaxter: Intentional change as part of this issue?

dvoytenko: Yeah. Intentional. Trying to scrub off as many globals as possible from the client-side code. It didn't seem like the `document.body` use case was strictly necessary, so I moved to the `baseElement` itself.

PR #363, opened 2019-03-14
Have innerHTML use document.createElement().

### LINKED ISSUE #350: Set innerHTML should use document.createElement
Should use `document.createElement[NS]` rather than `Element` constructor to make sure that correct HTMLElement subclasses are instantiated:

```ts
container.innerHTML = '<input>';
// Expected: true
// Actual: false
console.log(container.firstChild instanceof HTMLInputElement);
```

/to @delima02 /cc @jridgewell

delima02:
- Instead of using `new Element()` inside of `html-parser`, use `document.createElement()`
- Instead of using `new Text()` inside of `html-parser`, use `document.createTextNode()`
- Instead of using `new Comment()` inside of `html-parser`, use `document.createComment()`
- Fixes #350 

@jridgewell @choumx 
cc/ @caroqliu

### DIFF
--- a/src/test/element/innerHTML.ts
+++ b/src/test/element/innerHTML.ts
@@ -20,6 +20,7 @@ import { Text } from '../../worker-thread/dom/Text';
 import { Comment } from '../../worker-thread/dom/Comment';
 import { NodeType, SVG_NAMESPACE } from '../../transfer/TransferrableNodes';
 import { createDocument } from '../../worker-thread/dom/Document';
+import { HTMLInputElement } from '../../worker-thread/dom/HTMLInputElement';
 
 const test = anyTest as TestInterface<{
   node: Element;
@@ -178,6 +179,13 @@ test('set normalizes html namespace tag names', t => {
   const child = node.firstChild!;
   t.is(child.localName, 'div');
   t.is(child.nodeName, 'DIV');
+});
+
+test('set creates correct types of HTML elements', t => {
+  const { node } = t.context;
+  node.innerHTML = '<input>';
+  const child = node.firstChild!;
+  t.true(child instanceof HTMLInputElement);
 });
 
 test.skip('set has svg tags live in SVG namespace', t => {
--- a/src/third_party/html-parser/html-parser.ts
+++ b/src/third_party/html-parser/html-parser.ts
@@ -180,19 +180,15 @@ export function parse(data: string, rootElement: Element) {
   if (!valid) {
     throw new Error('Attempting to parse invalid HTML content.');
   }
+  
+  const child = root.firstChild;
 
-  root.childNodes.forEach((node: Node) => {
-    if (node instanceof Element) {
-      node.parentNode = null;
-    }
-  });
-
-  // remove the added <div>
-  if (root.firstChild) {
-    root.firstChild.childNodes.forEach((node: Node) => {
-      node.parentNode = null;
+  if (child) {
+    child.parentNode = null;
+    child.childNodes.forEach((node: Node) => {
+      node.parentNode = null;    
     });
-    return root.firstChild;
+    return child;
   }
 
   throw new Error('Attempting to parse invalid HTML.');
--- a/src/third_party/html-parser/html-parser.ts
+++ b/src/third_party/html-parser/html-parser.ts
@@ -180,15 +180,15 @@ export function parse(data: string, rootElement: Element) {
   if (!valid) {
     throw new Error('Attempting to parse invalid HTML content.');
   }
-  
-  const child = root.firstChild;
 
-  if (child) {
-    child.parentNode = null;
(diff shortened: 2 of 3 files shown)


src/third_party/html-parser/html-parser.ts
```diff
@@ -80,7 +78,8 @@ const kBlockTextElements: Elements = {
  */
 export function parse(data: string, rootElement: Element) {
   const ownerDocument = rootElement.ownerDocument;
-  const root = new Element(rootElement.nodeType, rootElement.nodeName, rootElement.namespaceURI, ownerDocument);
+  const root = 
+  ownerDocument.createElementNS(rootElement.namespaceURI, rootElement.localName);
```
choumx: Indentation is wrong. 

Looks like we don't use configure prettier to update code in `src/third_party/`? We should.

src/test/element/innerHTML.ts
```diff
@@ -180,6 +181,13 @@ test('set normalizes html namespace tag names', t => {
   t.is(child.nodeName, 'DIV');
 });
 
+test('set creates correct types of HTML elements', t => {
+  const { node } = t.context;
+  node.innerHTML = '<input>';
+  const child = node.firstChild!;
+  t.true(child instanceof HTMLInputElement);
+});
+
```
choumx: Add tests for comment and text nodes too.

src/third_party/html-parser/html-parser.ts
```diff
@@ -185,18 +181,16 @@ export function parse(data: string, rootElement: Element) {
     throw new Error('Attempting to parse invalid HTML content.');
   }
 
-  root.childNodes.forEach(node => {
+  root.childNodes.forEach((node: Node) => {
```
choumx: Tangent: Why do we need to iterate here? Shouldn't `root.childNodes.length == 1` due to the wrapping div?

src/third_party/html-parser/html-parser.ts
```diff
@@ -185,18 +181,16 @@ export function parse(data: string, rootElement: Element) {
...
   });
 
   // remove the added <div>
   if (root.firstChild) {
-    root.firstChild.childNodes.forEach(node => {
-      if (node instanceof Node) {
-        node.parentNode = null;
-      }
+    root.firstChild.childNodes.forEach((node: Node) => {
+      node.parentNode = null;
```
choumx: Tangent: This is a weird piece of code. Why do we need to disconnect the `parentNode` pointer for children of the wrapping div? It'll happen anyways in the `appendChild` calls in `set innerHTML`.

choumx:
- It's also not clear to me why we need to instantiate another element of the same type as `rootElement`. 
- `appendChild()` calls in `parse()` will trigger unnecessary `mutate()` calls.

src/worker-thread/dom/Element.ts
```diff
@@ -233,7 +233,7 @@ export class Element extends ParentNode {
     this.childNodes = [];
 
     // add new children
-    root.childNodes.forEach(n => {
+    root.childNodes.forEach((n: Node) => {
```
jridgewell: What type does this have without the cast?

src/third_party/html-parser/html-parser.ts
delima02: Correct. Thanks!

src/third_party/html-parser/html-parser.ts
delima02: This is actually needed for non-nested nodes. The original logic returns a single node (which is sort of an 'invisible' node, it only really returns its children). If we don't do this disconnect, an array of nodes must be returned instead (`root.firstChild.childNodes`). Otherwise only the first child will be returned.

src/worker-thread/dom/Element.ts
delima02: Before the bug fix, `Node`. After, `any`. I'm not entirely sure why.

src/third_party/html-parser/html-parser.ts
```diff
@@ -180,19 +180,15 @@ export function parse(data: string, rootElement: Element) {
   if (!valid) {
     throw new Error('Attempting to parse invalid HTML content.');
   }
+  
+  const child = root.firstChild;
```
choumx: Nit: `const wrapper`

src/test/element/innerHTML.ts
delima02: As discussed in person: creating text and comment nodes has an equal result when using the constructor and the `document.create...` method. Testing for actual method calls will be addressed in a separate PR.

src/third_party/html-parser/html-parser.ts
delima02: Per our discussion, separate PR.

PR #365, opened 2019-03-14
Update web_compat_table to reflect current support

nainar:
Added `getBoundingClientRect()` and setter for `innerHTML`

### DIFF
--- a/web_compat_table.md
+++ b/web_compat_table.md
@@ -196,7 +196,7 @@ This section highlights the DOM APIs that are implemented in WorkerDOM currently
 | Element.getAttributeNode()                          | ✖️     |                                                  | 
 | Element.getAttributeNodeNS()                        | ✖️     |                                                  | 
 | Element.getAttributeNS()                            | ✔️     |                                                  | 
-| Element.getBoundingClientRect()                     | ✔️     |                                                  | 
+| Element.getBoundingClientRect()                     | ✔️(Async)| Implemented as Element.getBoundingClientRectAsync()                                                 | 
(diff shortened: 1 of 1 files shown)


web_compat_table.md
```diff
@@ -196,7 +196,7 @@ This section highlights the DOM APIs that are implemented in WorkerDOM currently
 | Element.getAttributeNode()                          | ✖️     |                                                  | 
 | Element.getAttributeNodeNS()                        | ✖️     |                                                  | 
 | Element.getAttributeNS()                            | ✔️     |                                                  | 
-| Element.getBoundingClientRect()                     | ✖️     |                                                  | 
+| Element.getBoundingClientRect()                     | ✔️     |                                                  |
```
choumx: I believe this is only available in async form: `getBoundingClientRectAsync()`. If so, keep as "X" and add a note?

nainar: Do we want to add that to the notes section. 

Element.getBoundingClientRect()  | ✔️ | Implemented as an aysnc function getBoundingClientRectAsync()

kristoferbaxter: Since this will be a frequent occurrence, a sync method is replaced with an async equivalent, can we mark this with a significant identifier that's reusable?

nainar: Sorry I missed this @kristoferbaxter does it look good now?

CodeWithOz:
Hi, please what's the status of `getBoundingClientRectAsync`? #259 suggests that this available but it seems that a decision was made here to not update the docs to reflect this.

PR #368, opened 2019-03-15
Have set innerHTML handle SVG namespace tags.

delima02:
- The `<svg>` tag now lives in the SVG namespace.
- All child nodes of an `<svg>` tag now live in the SVG namespace.
- Child nodes of a `<foreignObject>` tag will live in the HTML namespace, but the tag itself lives in the SVG namespace.

@choumx @jridgewell


src/third_party/html-parser/html-parser.ts
```diff
@@ -110,15 +112,31 @@ export function parse(data: string, rootElement: Element) {
 
     const normalizedTagName = tagName.toUpperCase();
 
+    if (normalizedTagName === 'SVG') {
+      if (beginningSlash) {
+        currentNamespace = HTML_NAMESPACE;
+      } else currentNamespace = SVG_NAMESPACE;
```
jridgewell: If else statements needs curly braces. You can use a ternary if you want to be more concise.

src/third_party/html-parser/html-parser.ts
```diff
@@ -140,6 +158,13 @@ export function parse(data: string, rootElement: Element) {
         }
       }
     }
+
+    if (tagName === 'foreignObject' && currentNamespace === SVG_NAMESPACE) {
+      if (beginningSlash) {
+        currentNamespace = SVG_NAMESPACE;
+      } else currentNamespace = HTML_NAMESPACE;
```
jridgewell: This will create SVG elements in the HTML namespace:

```js
div.innerHTML = `
  <svg>
    <foreignObject></foreignObject>
    <feImage></feImage>
  </svg>
`;

File: src/third_party/html-parser/html-parser.ts [hunk 3]
```diff
@@ -110,15 +112,29 @@ export function parse(data: string, rootElement: Element) {
 
     const normalizedTagName = tagName.toUpperCase();
 
+    if (normalizedTagName === 'SVG') {
+      beginningSlash ? currentNamespace = HTML_NAMESPACE : currentNamespace = SVG_NAMESPACE;
```
kristoferbaxter: Here, we can use the ternary only on the value side of this setter `currentNamespace = beginningSlash ? HTML_NAMESPACE : SVG_NAMESPACE`.

src/third_party/html-parser/html-parser.ts
```diff
@@ -110,15 +112,29 @@ export function parse(data: string, rootElement: Element) {
...
       // not </ tags
       if (!endSlash && kElementsClosedByOpening[currentParent.tagName]) {
         if (kElementsClosedByOpening[currentParent.tagName][normalizedTagName]) {
           tagsClosed.push(currentParent.tagName);
         }
       }
-      const childToAppend = 
-      ownerDocument.createElementNS(currentParent.namespaceURI, tagName.toLowerCase());
+
+      let childToAppend;
```
kristoferbaxter: `const childToAppend = ownerDocument.createElementNS(currentNamespace, currentNamespace === SVG_NAMESPACE ? tagName : toLower(tagName))`

`toLower` can be imported from utils.

`WorkerDOM` doesn't throw errors right now, so the error case can be done seperately from the `const` setter.

src/third_party/html-parser/html-parser.ts
```diff
@@ -140,6 +156,11 @@ export function parse(data: string, rootElement: Element) {
         }
       }
     }
+
+    if (tagName === 'foreignObject') {
+      beginningSlash ? currentNamespace = SVG_NAMESPACE : currentNamespace = HTML_NAMESPACE;
```
kristoferbaxter: Same deal here with the ternary on the value side of the setter.

src/third_party/html-parser/html-parser.ts
delima02: Thanks!

src/third_party/html-parser/html-parser.ts
```diff
@@ -110,15 +113,24 @@ export function parse(data: string, rootElement: Element) {
...
       if (!endSlash && kElementsClosedByOpening[currentParent.tagName]) {
         if (kElementsClosedByOpening[currentParent.tagName][normalizedTagName]) {
           tagsClosed.push(currentParent.tagName);
         }
       }
-      const childToAppend = 
-      ownerDocument.createElementNS(currentParent.namespaceURI, tagName.toLowerCase());
+
+      if (currentNamespace !== SVG_NAMESPACE && currentNamespace !== HTML_NAMESPACE) {
+        throw new Error("Namespace not supported.");
```
jridgewell: I believe this is only possible from the root element's namespace, so let's move this check up there.

src/third_party/html-parser/html-parser.ts
```diff
@@ -110,15 +113,24 @@ export function parse(data: string, rootElement: Element) {
...
+      }
+
+      const childToAppend = ownerDocument.createElementNS(
+        currentNamespace, currentNamespace === SVG_NAMESPACE ? tagName : toLower(tagName));
```
jridgewell: Nit: the MATHML_NAMESPACE also uses XML camel-case rules. Even if we don't support it yet, we can minimize necessary changes if we ever do by only lowercasing for the HTML namespace.

src/third_party/html-parser/html-parser.ts
```diff
@@ -140,6 +152,11 @@ export function parse(data: string, rootElement: Element) {
         }
       }
     }
+
+    if (tagName === 'foreignObject') {
+      currentNamespace = beginningSlash ? SVG_NAMESPACE : HTML_NAMESPACE;
```
jridgewell: Please add a test for sibling elements after a `<foreignObject>`.

src/third_party/html-parser/html-parser.ts
```diff
@@ -82,12 +84,17 @@ export function parse(data: string, rootElement: Element) {
...
   let currentParent = root as Node;
+  let currentNamespace = root.namespaceURI;
   const stack = [root as Node];
   let lastTextPos = 0;
   let match: RegExpExecArray | null;
   data = '<div>' + data + '</div>';
   const tagsClosed = [] as string[];
 
+  if (currentNamespace !== SVG_NAMESPACE && currentNamespace !== HTML_NAMESPACE) {
+    throw new Error("Namespace not supported.");
```
choumx: Include the unsupported namespace string in the error.

src/test/element/innerHTML.ts
```diff
@@ -188,17 +188,48 @@ test('set creates correct types of HTML elements', t => {
...
+  t.is(foreignObjectChild.namespaceURI, HTML_NAMESPACE);
+});
+
+test('set resets namespace after foreignObject end tag', t => {
+  const { node } = t.context;
+  node.innerHTML = '<svg><foreignObject></foreignObject><div></div></svg>';
+
+  const svgWrapper = node.firstChild!;
+  const foreignObjectSibling = svgWrapper.lastChild!; // div node
+  t.is(foreignObjectSibling.namespaceURI, SVG_NAMESPACE);
```
choumx: Add test for unsupported namespace case e.g. MathML.

choumx:
Please also fix CI builds.

> FAIL  ./dist/worker.safe.mjs: 8.63KB > maxSize 8.6KB (brotli)

PR #375, opened 2019-03-19
Change the html-parser's wrapping element.

### LINKED ISSUE #372: innerHTML doesn't work for modifying <p> tags
```
// <p id="myParagraph"></p>
document.getElementById('myParagraph').innerHTML = 'hello';
```

This throws `Attempting to parse invalid HTML content.` - I believe that's because of this bit:

https://github.com/ampproject/worker-dom/blob/0333ae5bef9961a145690e153102d8d98eb05f72/src/third_party/html-parser/html-parser.ts#L88

Since `<div>` is not allowed inside `<p>`, this produces invalid HTML.

delima02:
- When parsing html, change the element wrapping the content from a `<div>` to a `<q>`, which will not close any other tag automatically.
- Fixes #372 
- Fix test that failed after changing this behavior (was wrongly phrased in the first place).

@choumx @jridgewell 
cc/ @fstanis

### DIFF
--- a/src/test/element/innerHTML.ts
+++ b/src/test/element/innerHTML.ts
@@ -172,9 +172,8 @@ test('set will alter root element\'s contents, not the element itself', t => {
   const document = createDocument();
   const pNode = document.createElement('p');
 
-  const testString = 'Hello World!';
-  pNode.innerHTML = testString;
-  t.is(pNode.textContent, testString);
+  pNode.innerHTML = 'Hello World!';
+  t.is(pNode.textContent, 'Hello World!');
 });
 
 test('set takes all block text element content as text', t => {
--- a/src/test/element/innerHTML.ts
+++ b/src/test/element/innerHTML.ts
@@ -159,10 +159,13 @@ test('set closes tags by closing others', t => {
   t.is(child.nodeName, 'DIV');
   t.is(child.firstChild!.nodeName, 'A');
 
-  node.innerHTML = '<a><div></div>';
-  child = node.firstChild!;
-  t.is(child.nodeName, 'A');
-  t.is(child.firstChild!.nodeName, 'DIV');
+  node.innerHTML = '<p><div></div>';
+  const pChild = node.firstChild!;
+  const divChild = node.lastChild!;
+  t.true(node.childNodes.length === 2);
+  t.is(pChild.nodeName, 'P');
+  t.is(divChild.nodeName, 'DIV');
+  
 });
 
 test('set takes all block text element content as text', t => {
--- a/src/test/element/innerHTML.ts
+++ b/src/test/element/innerHTML.ts
@@ -160,11 +160,8 @@ test('set closes tags by closing others', t => {
   t.is(child.firstChild!.nodeName, 'A');
 
   node.innerHTML = '<p><div></div>';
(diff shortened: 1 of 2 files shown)


src/test/element/innerHTML.ts
```diff
@@ -159,10 +159,13 @@ test('set closes tags by closing others', t => {
...
-  node.innerHTML = '<a><div></div>';
-  child = node.firstChild!;
-  t.is(child.nodeName, 'A');
-  t.is(child.firstChild!.nodeName, 'DIV');
+  node.innerHTML = '<p><div></div>';
+  const pChild = node.firstChild!;
+  const divChild = node.lastChild!;
+  t.true(node.childNodes.length === 2);
+  t.is(pChild.nodeName, 'P');
+  t.is(divChild.nodeName, 'DIV');
```
choumx: Let's also add a new test for the case referenced in the bug. Add a comment above the new test case referencing the issue number.

src/test/element/innerHTML.ts
```diff
@@ -159,10 +159,24 @@ test('set closes tags by closing others', t => {
...
```
choumx:
```ts
t.is(node.innerHTML, '<p></p><div></div>');
```

src/test/element/innerHTML.ts
```diff
@@ -159,10 +159,24 @@ test('set closes tags by closing others', t => {
...
+
+// Expected behavior, as discussed in:
+// https://github.com/ampproject/worker-dom/issues/372
+test('set will not alter root element', t => {
+  const document = createDocument();
+  const pNode = document.createElement('p');
+  // normally, a <div> would automatically close a <p>
+  pNode.innerHTML = '<div></div>';
+  const child = pNode.firstChild!;
+  t.is(child.nodeName, 'DIV');
```
choumx: This doesn't exactly cover the referenced issue. You should just try setting the innerHTML of a `<p>` to some text.

src/test/element/innerHTML.ts
```diff
@@ -159,10 +159,24 @@ test('set closes tags by closing others', t => {
...
+  node.innerHTML = '<p><div></div>';
+  const pChild = node.firstChild!;
+  const divChild = node.lastChild!;
+  t.true(node.childNodes.length === 2);
+  t.is(pChild.nodeName, 'P');
+  t.is(divChild.nodeName, 'DIV');
+  
+});
+
+// Expected behavior, as discussed in:
```
choumx: This comment can be better.

src/test/element/innerHTML.ts
```diff
@@ -160,23 +160,21 @@ test('set closes tags by closing others', t => {
...
   const document = createDocument();
   const pNode = document.createElement('p');
-  // normally, a <div> would automatically close a <p>
-  pNode.innerHTML = '<div></div>';
-  const child = pNode.firstChild!;
-  t.is(child.nodeName, 'DIV');
+
+  const testString = 'Hello World!';
+  pNode.innerHTML = testString;
+  t.is(pNode.textContent, testString);
```
choumx: Nit: Prefer duplicating literals to make test code easier to read.

`go/tott-510`

PR #380, opened 2019-03-21
Implement Canvas API

delima02:
- Implementation of the Canvas API in the worker thread.
- This will work for Chrome and FireFox (the latter if the user explicitly enables `OffscreenCanvas`).
- 04/16/2019: OffscreenCanvas Polyfill has been added to this PR. Most `CanvasRenderingContext2D` methods will now work on all browsers.

@choumx @jridgewell


kristoferbaxter:
Going to request this change holds off until the buffer branch lands since it will change the messaging format.

I'll review the changes tomorrow.

src/test/htmlcanvaselement/context.ts
```diff
@@ -0,0 +1,2539 @@
...
+    });
+    
+    test('context only calls upgraded clearRect if available', async t => {
+        const { context2d, deferredUpgrade } = t.context;
+
+        const instance = new OffscreenCanvas();
+        const stub = createStub(instance.getContext('2d'), "clearRect");
+        const promise = new Promise((res) => {
+            deferredUpgrade.resolve(instance);
+            res();
```
jridgewell: While this technically works, a better practice it to await the _same_ promise instance that is needed by the code, not another instance that happens to resolve afterwards.

src/test/htmlcanvaselement/context.ts
```diff
@@ -0,0 +1,2539 @@
...
+        });
+        promise.then(() => {
```
jridgewell: This needs to be awaited on, to ensure that the then callback has run (and may throw) before the test completes.

For instance, if you put in a known assertion inside this then block, it's not going to fail the overall test.

src/test/htmlcanvaselement/context.ts
```diff
@@ -0,0 +1,2539 @@
...
+            context2d.clearRect(1, 2, 3, 4);
+            t.true(stub.withArgs(1, 2, 3, 4).calledOnce);
+        });
+    });
+
+    test('context calls both versions of clearRect when called before upgrade', async t => {
```
jridgewell: This is already tested by the first two tests. What's needed here is that it only calls the upgraded instance if it is already upgraded.

src/test/htmlcanvaselement/context.ts
```diff
@@ -0,0 +1,2539 @@
...
+            deferredUpgrade.resolve(instance);
+            res();
+        });
+        promise.then(() => {
+            t.true(instanceSpy.withArgs(200).calledOnce);
+        });
+    });
+});
+
+// METHODS
```
jridgewell: Hm?

src/test/htmlcanvaselement/context.ts
```diff
@@ -0,0 +1,2539 @@
...
+        });
+    });
+});
+
+// SETTERS
+describe('lineWidth', () => {
+    test('context calls set lineWidth', t => {
+        const { context2d, implementation } = t.context;
+        const spy = sandbox.spy();
+        createSetterStub(implementation, "lineWidth", spy);
```
jridgewell: This is a different API than `createStub`. Is there a reason to require a spy/stub to be passed in?

src/test/tsconfig.json
```diff
@@ -12,14 +12,16 @@
...
   },
   "include": [
-    "**/*.ts"
+    "**/*.ts",
+    "**/*.js"
   ],
   "exclude": [
-    "main-thread/**/*.ts"
+    "main-thread/**/*.ts",
+    "node_modules/canvas"
```
jridgewell: Undo

src/test/tsconfig.json
```diff
@@ -12,14 +12,16 @@
...
     "strictNullChecks": true,
-    "noImplicitAny": true,
+    "noImplicitAny": false,
     "noImplicitThis": true,
     "alwaysStrict": true
   },
   "include": [
-    "**/*.ts"
+    "**/*.ts",
+    "**/*.js"
```
jridgewell: I don't think we need this anymore.

src/test/tsconfig.json
```diff
@@ -12,14 +12,16 @@
     "allowJs": false,
     "noUnusedLocals": true,
     "strictNullChecks": true,
-    "noImplicitAny": true,
+    "noImplicitAny": false,
```
jridgewell: Is this still needed?

src/test/htmlcanvaselement/context.ts
```diff
@@ -0,0 +1,2539 @@
...
+    sandbox = sinon.createSandbox();
+    const document = createDocument();
+    const canvas = document.createElement('canvas') as HTMLCanvasElement;
+
+    t.context = {
+        canvas,
+        context2d: canvas.getContext('2d'),
+        deferredUpgrade: deferredUpgrades.get(canvas),
+        implementation: OffscreenCanvas.implementation,
+        image: document.createElement('img') as HTMLImageElement,
```
jridgewell: Is this needed in the context? Can't this just be created in the appropriate tests?

src/test/htmlcanvaselement/context.ts
delima02: A `stub` does not detect method calls for setters. I'm not sure what the reason for this is, but I looked it up and it just seems their accepted behavior

src/test/htmlcanvaselement/context.ts
delima02: I think I just addressed this for `clearRect`, not quite sure if that's what you meant. Please LMK if so, so I can change all other tests.

src/test/htmlcanvaselement/context.ts
```diff
@@ -0,0 +1,2539 @@
...
+    context2d: CanvasRenderingContext2D;
+    deferredUpgrade: {resolve: (instance: OffscreenCanvas) => void, reject: (errorMsg: string) => void};
+    implementation: CanvasRenderingContext2DImplementation;
+    image: HTMLImageElement;
+}>;
+
+let sandbox: sinon.SinonSandbox;
+
+class OffscreenCanvas {
+    public static implementation: CanvasRenderingContext2DImplementation;
```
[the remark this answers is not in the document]
delima02: Good idea :)

src/test/htmlcanvaselement/context.ts
delima02: Just did this for simplicity. Since an `image` element is needed by multiple tests and the `document` element (which is already here) is needed to create one. I can move it to the individual tests, but it will take more lines of code. Should I?

src/test/htmlcanvaselement/context.ts
delima02: I was doing some research and it doesn't seem common to await a `then` hook from out of scope (as we need here). This is why I used the extra promise. I tried a few approaches for this (such as passing in a callback to this `resolve`), but none seemed to work appropriately. I'd be happy to discuss this in person.

src/test/htmlcanvaselement/context.ts
delima02: Never mind, found a way of not even needing it.

delima02:
@jridgewell ready for review 👍

src/worker-thread/dom/HTMLCanvasElement.ts
```diff
@@ -0,0 +1,44 @@
...
+
+import { HTMLElement } from './HTMLElement';
+import { registerSubclass } from './Element';
+import { reflectProperties } from './enhanceElement';
+import { CanvasRenderingContext2D } from '../CanvasRenderingContext2D';
+
+export class HTMLCanvasElement extends HTMLElement {
+    getContext(contextType: string): CanvasRenderingContext2D {
+        if (contextType == '2D' || contextType == '2d') {
+            return new CanvasRenderingContext2D(this);
```
choumx: Is `getContext()` supposed to instantiate a new context on each invocation?

src/worker-thread/CanvasRenderingContext2D.ts
```diff
@@ -0,0 +1,376 @@
...
+export const deferredUpgrades = new WeakMap();
+
+export function getOffscreenCanvasAsync(canvas: HTMLCanvasElement): Promise<{getContext(c: '2d'): CanvasRenderingContext2DImplementation}> {
+    return new Promise((resolve, reject) => {
+        // TODO: This should only happen in test environemnet. Otherwise, we should throw.
+        if (typeof addEventListener !== 'function') {
+            const deferred = { resolve, reject };
+            deferredUpgrades.set(canvas, deferred);
+        } else {
+            addEventListener('message', ({ data }: { data: MessageToWorker }) => {
```
choumx: Can we remove the event listener once the offscreen canvas is transferred?

src/worker-thread/CanvasRenderingContext2D.ts
```diff
@@ -0,0 +1,376 @@
...
+                    
+                } else {
+                    (this.implementation as any)[call.fnName](...call.args);
+                }
+            }
+
+        });
+    }
+
+    private delegate(fnName: string, fnArgs: any[], isSetter: boolean) {
```
choumx: Did Justin explain the `Object.defineProperty` technique yet to hopefully avoid a lot of this boilerplate?

src/worker-thread/CanvasRenderingContext2D.ts
```diff
@@ -0,0 +1,376 @@
...
+                if (data[TransferrableKeys.type] === MessageType.OFFSCREEN_CANVAS_INSTANCE) {
+                    const transferredOffscreenCanvas = (data as OffscreenCanvasToWorker)[TransferrableKeys.data];
+                    resolve(transferredOffscreenCanvas as {getContext(c: '2d'): CanvasRenderingContext2DImplementation});
+                }
+            });
+        }
+    })
+  }
+
+export class CanvasRenderingContext2D implements CanvasRenderingContext2DImplementation {
```
choumx: This seems backwards. The interface should not be called "implementation", the class that implements it should be.

src/worker-thread/dom/HTMLCanvasElement.ts
delima02: Yes! Just had not pushed this yet :)

delima02:
@choumx I added a simple demo to this PR, called Canvas. Let me know if you have any questions on using it / the control flow :)

src/worker-thread/CanvasRenderingContext2D.ts
delima02: hmm not really, I will ask him about this next time I see him. Thanks!

src/main-thread/commands/offscreen-canvas.ts
```diff
@@ -0,0 +1,30 @@
+import { WorkerContext } from "../worker";
+import { TransferrableKeys } from "../../transfer/TransferrableKeys";
+import { MessageType } from "../../transfer/Messages";
+import { CommandExecutor } from "./interface";
+import { OffscreenCanvasMutationIndex } from "../../transfer/TransferrableMutation";
+
+export function OffscreenCanvasProcessor(workerContext: WorkerContext): CommandExecutor {
+    return {
```
jridgewell: The whitespace indentation in this file should be 2 spaces, right?

src/test/htmlcanvaselement/context.ts
```diff
@@ -0,0 +1,2611 @@
...
+        const stub = createStub(implementation, "clearRect");
+        context2d.clearRect(1, 2, 3, 4);
+        t.true(stub.withArgs(1, 2, 3, 4).calledOnce);
+    });
+    
+    test('context only calls upgraded clearRect if available', async t => {
+        const { context2d, deferredUpgrade, implementation } = t.context;
+
+        const instance = new OffscreenCanvas();
+        const stub = createStub(instance.getContext('2d'), "clearRect");
```
jridgewell: Use a single consistent quote, unless there's a good reason to change to a different quote.

src/test/htmlcanvaselement/reflectedProperties.ts
```diff
@@ -0,0 +1,35 @@
...
+
+test.beforeEach(t => {
+  const document = createDocument();
+
+  t.context = {
+    element: document.createElement('canvas') as HTMLCanvasElement,
+  };
+});
+
+testReflectedProperty({ width: [0] });
```
jridgewell: Add tests for the `getContext` method.

src/transfer/Messages.ts
```diff
@@ -64,4 +65,9 @@ export interface BoundingClientRectToWorker {
   [TransferrableKeys.target]: TransferredNode;
   [TransferrableKeys.data]: TransferrableBoundingClientRect;
 }
-export type MessageToWorker = EventToWorker | ValueSyncToWorker | BoundingClientRectToWorker;
+export interface OffscreenCanvasToWorker {
+  [TransferrableKeys.type]: MessageType.OFFSCREEN_CANVAS_INSTANCE;
+  [TransferrableKeys.target]: TransferredNode;
+  [TransferrableKeys.data]: Object; // This will be an OffscreenCanvas
```
jridgewell: I think we changed this to a `Transferrable`?

src/worker-thread/CanvasRenderingContext2D.ts
```diff
@@ -0,0 +1,382 @@
...
+import { TransferrableKeys } from '../transfer/TransferrableKeys';
+import { transfer } from './MutationTransfer';
+import { TransferrableMutationType } from '../transfer/TransferrableMutation';
+
+declare var OffscreenCanvas: any;
+
+export const deferredUpgrades = new WeakMap();
+
+export function getOffscreenCanvasAsync(canvas: HTMLCanvasElement): Promise<{getContext(c: '2d'): CanvasRenderingContext2D}> {
+    return new Promise((resolve, reject) => {
```
jridgewell: Ditto whitespace.

src/worker-thread/CanvasRenderingContext2D.ts
```diff
@@ -0,0 +1,382 @@
...
+        const messageHandler = ({ data }: { data: MessageToWorker }) => {
+            if (data[TransferrableKeys.type] === MessageType.OFFSCREEN_CANVAS_INSTANCE) {
```
jridgewell: Add a new test case that creates two canvas elements in the worker thread, and gets tries to draw the houses on them.

src/worker-thread/CanvasRenderingContext2D.ts
```diff
@@ -0,0 +1,382 @@
...
+    constructor(canvas: HTMLCanvasElement) {
+        this.canvasElement = canvas;
+
+        this.goodOffscreenPromise = getOffscreenCanvasAsync(this.canvasElement).then((instance) => {
+            this.implementation = instance.getContext('2d');
+            this.upgraded = true;
+
+            for (const call of this.calls) {
+                if (call.setter) {
+                    if (call.args.length != 1) {
```
jridgewell: I don't believe this can ever happen.

src/worker-thread/CanvasRenderingContext2D.ts
jridgewell: If we do `Object.defineProperty` to create all the methods, I think we'll loose the typing information on the object.

src/worker-thread/CanvasRenderingContext2D.ts
```diff
@@ -0,0 +1,382 @@
...
+                        throw new Error(
+                            "Attempting to set property with wrong number of arguments.");
+                    }
+                    (this.implementation as any)[call.fnName] = call.args[0];
+                    
+                } else {
+                    (this.implementation as any)[call.fnName](...call.args);
+                }
+            }
+
```
jridgewell: We have a memory leak here. Can you find it?

src/transfer/Messages.ts
delima02: We do that in `src/main-thread/main-thread.d.ts`. I don't think it'll work here because `Transferable` does not have method `getContext` (as it'll be required in `CanvasRenderingContext2DImplementation`'s initialization).

src/worker-thread/CanvasRenderingContext2D.ts
delima02: Just being cautious. Should I remove?

src/main-thread/commands/offscreen-canvas.ts
kristoferbaxter: Prettier should be taking care of these items, is it not running against your commits for some reason?

src/test/htmlcanvaselement/context.ts
kristoferbaxter: Same feedback here about Prettier.

src/main-thread/commands/offscreen-canvas.ts
delima02: Commit hooks weren't running because Santa was blocking them on my computer. Fixed that.

src/worker-thread/CanvasRenderingContext2D.ts
delima02: I tried this locally and it will create two canvas elements, yet will only draw on one. Will be thinking about why this happens as well as how to test this behavior.

src/main-thread/commands/offscreen-canvas.ts
kristoferbaxter: Good to know, we might need to invest in a bot for this since they're not guaranteed on each person's machines.

delima02:
Ready from my end! Final comments??

src/worker-thread/CanvasRenderingContext2D.ts
```diff
@@ -0,0 +1,381 @@
...
+    for (const call of this.calls) {
+      if (call.setter) {
+        if (call.args.length != 1) {
+          throw new Error('Attempting to set property with wrong number of arguments.');
+        }
+        (this.implementation as any)[call.fnName] = call.args[0];
+      } else {
+        (this.implementation as any)[call.fnName](...call.args);
+      }
+    }
```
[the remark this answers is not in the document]
delima02: Missed it when making this a separate method.

src/main-thread/worker.ts
```diff
@@ -93,13 +93,13 @@ export class WorkerContext {
...
-  messageToWorker(message: MessageToWorker) {
+  messageToWorker(message: MessageToWorker, transferables: Transferable[]) {
     if (DEBUG_ENABLED) {
       console.info('debug', 'messageToWorker', readableMessageToWorker(this.nodeContext, message));
     }
     if (this.config.onSendMessage) {
       this.config.onSendMessage(message);
     }
-    this.worker.postMessage(message);
+    this.worker.postMessage(message, transferables);
```
choumx: Should this be an optional parameter instead of forcing all messages without transferables to pass an empty array?

src/test/htmlcanvaselement/context.test.ts
```diff
@@ -0,0 +1,2608 @@
...
+
+    deferredUpgrade.resolve(instance);
+
+    await context2d.goodOffscreenPromise.then(() => {
+      t.true(instanceStub.withArgs(0, 1).calledOnce);
+    });
+  });
+});
+
+describe('rota
```
choumx: Any reason why we're not using the more conventional form inline?

```js
const myStub = sandbox.stub(obj, method);
```
https://sinonjs.org/releases/latest/stubs/

src/test/htmlcanvaselement/context.test.ts
```diff
@@ -0,0 +1,2608 @@
...
+  t.context = {
+    canvas,
+    context2d: canvas.getContext('2d'),
+    deferredUpgrade: deferredUpgrades.get(canvas),
+    implementation: OffscreenCanvas.mostRecentInstance.getContext('2d'),
+  };
+});
+
+describe('clearRect', () => {
+  test('context calls clearRect', t => {
```
choumx: Using a fake `describe()` is mixing testing paradigms. Better to just use comment blocks and stick to Ava style.

Discussion thread if you're curious: https://github.com/avajs/ava/issues/222

src/test/htmlcanvaselement/context.test.ts
choumx: s/spy/setter

A "spy" isn't the right name to describe this function.

choumx: :)

src/worker-thread/CanvasRenderingContext2D.ts
```diff
@@ -0,0 +1,371 @@
...
+        data[TransferrableKeys.type] === MessageType.OFFSCREEN_CANVAS_INSTANCE &&
+        data[TransferrableKeys.target][0] === canvas[TransferrableKeys.index]
+      ) {
+        removeEventListener('message', messageHandler);
+        const transferredOffscreenCanvas = (data as OffscreenCanvasToWorker)[TransferrableKeys.data];
+        resolve(transferredOffscreenCanvas as { getContext(c: '2d'): CanvasRenderingContext2D });
+      }
+    };
+
+    // TODO: This should only happen in test environemnet. Otherwise, we should throw.
```
choumx: Fix and remove TODOs.

src/worker-thread/CanvasRenderingContext2D.ts
```diff
@@ -0,0 +1,371 @@
...
+  /* THE CANVAS STATE */
+  save(): void {
+    this.delegate('save', [...arguments], false);
+  }
+
+  restore(): void {
+    this.delegate('restore', [...arguments], false);
+  }
+
+  // canvas property is readonly. We don't want to implement getters, but this must be here
```
choumx: > We don't want to implement getters

What does this mean?

src/worker-thread/DOMTypes.ts
```diff
@@ -0,0 +1,205 @@
+import { HTMLElement } from './dom/HTMLElement';
```
choumx: I think Canvas interfaces are technically not part of DOM, so DOMTypes.ts doesn't seem like the right name. Also, if there's no compilable code this should probably be a `d.ts` declaration file: https://www.typescriptlang.org/docs/handbook/declaration-files/by-example.html

Also, let's group these files under a new folder `src/worker-thread/canvas`.

src/main-thread/commands/offscreen-polyfill-calls.ts
```diff
@@ -0,0 +1,103 @@
+import { CommandExecutor } from './interface';
+import { WorkerContext } from '../worker';
+import { OffscreenContextPolyfillMutationIndex } from '../../transfer/TransferrableMutation';
+import { NumericBoolean } from '../../utils';
+import { Strings } from '../strings';
+
+export function OffscreenPolyfillCallProcessor(strings: Strings, workerContext: WorkerContext): CommandExecutor {
+  return {
+    execute(mutations: Uint16Array, startPosition: number, target: RenderableElement): number {
```
choumx: Do we have tests that cover these processor functions?

src/test/htmlcanvaselement/context.test.ts
delima02: Because we use a fake context object (`method` does not actually exist on this `obj`).

src/worker-thread/CanvasRenderingContext2D.ts
delima02: Will include this in a later PR (there is not a simple way to do this currently, and it does not affect the operation of this API outside of testing environment)

src/test/htmlcanvaselement/context.test.ts
delima02: thanks!

src/main-thread/commands/offscreen-polyfill-calls.ts
delima02: Now we do 👍

src/test/htmlcanvaselement/context.test.ts
```diff
@@ -0,0 +1,2608 @@
...
+    const context = ({} as unknown) as CanvasRenderingContext2D;
+    OffscreenCanvas.mostRecentInstance = this;
+    this.implementation = context;
+  }
+
+  getContext(c: string): CanvasRenderingContext2D {
+    return this.implementation;
+  }
+}
+(global as any).OffscreenCanvas = OffscreenCanvas;
```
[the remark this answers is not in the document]
delima02: Does this new approach seem like a good idea?

src/test/htmlcanvaselement/context.test.ts
```diff
@@ -0,0 +1,2608 @@
...
+let sandbox: sinon.SinonSandbox;
+
+class OffscreenCanvas {
+  public static mostRecentInstance: OffscreenCanvas;
+  public implementation: CanvasRenderingContext2D;
+
+  constructor() {
+    // this.x, y
+    const context = ({} as unknown) as CanvasRenderingContext2D;
+    OffscreenCanvas.mostRecentInstance = this;
```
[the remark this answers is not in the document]
delima02: From our discussion yesterday: this is actually needed for calls that happen before the upgraded (auto-synchronized) version of `OffscreenCanvas` is fetched from the main-thread.

src/test/htmlcanvaselement/context.test.ts
```diff
@@ -0,0 +1,3068 @@
...
+    deferredUpgrade: deferredUpgrades.get(canvas),
+    implementation: FakeOffscreenCanvas.mostRecentInstance.getContext('2d'),
+    sandbox,
+  };
+});
+
+test.afterEach(t => {
+  const { sandbox } = t.context;
+  sandbox.restore();
+  sandbox.verify();
```
choumx: This verifies mocks. Do we have any mocks?

src/test/htmlcanvaselement/context.test.ts
```diff
@@ -0,0 +1,3068 @@
...
+
+class FakeOffscreenCanvas {
+  // For testing convenience; this class looks different from the OffscreenCanvas spec API
+  public static mostRecentInstance: FakeOffscreenCanvas;
+  public implementation: CanvasRenderingContext2D;
+
+  constructor() {
+    // this.x, y
+    const context = ({} as unknown) as CanvasRenderingContext2D;
+    FakeOffscreenCanvas.mostRecentInstance = this;
```
choumx: This is still an issue. Remember what we discussed was the possible solution for this?

src/test/DocumentCreation.ts
```diff
@@ -117,8 +119,9 @@ const GlobalScope: GlobalScope = {
   HTMLTimeElement,
 };
 
-export function createTestingDocument(): Document {
-  const document = new Document(GlobalScope);
+export function createTestingDocument(additional: {} | null = null): Document {
```
choumx: s/additional/overrides

choumx: Add JSDoc with `@param` describing how to use this new parameter.

src/test/htmlcanvaselement/reflectedProperties.test.ts
```diff
@@ -0,0 +1,64 @@
...
+    element: document.createElement('canvas') as HTMLCanvasElement,
+  };
+});
+
+class OffscreenCanvas {
+  getContext(c: string): CanvasRenderingContext2D {
+    return ({} as unknown) as CanvasRenderingContext2D;
+  }
+}
+(global as any).OffscreenCanvas = OffscreenCanvas;
```
choumx: Same problem.

src/test/main-thread/offscreen-canvas.test.ts
```diff
@@ -0,0 +1,258 @@
...
+
+function createStub(obj: any, method: string) {
+  obj[method] = sandbox.stub();
+  return obj[method];
+}
+
+function createSetterStub(obj: any, property: string, spy: () => {}) {
+  obj[property] = 'existent';
+  sandbox.stub(obj, property).set(spy);
+}
```
choumx: Same here.

src/worker-thread/canvas/OffscreenCanvasPolyfill.ts
```diff
@@ -0,0 +1,311 @@
...
+  CanvasDirection,
+  CanvasFillRule,
+} from './CanvasTypes';
+import { transfer } from '../MutationTransfer';
+import { Document } from '../dom/Document';
+import { NumericBoolean } from '../../utils';
+import { store } from '../strings';
+import { HTMLElement } from '../dom/HTMLElement';
+
+export class OffscreenCanvasPolyfill<ElementType extends HTMLElement> {
```
choumx: At least write JSDoc for each class.

src/test/htmlcanvaselement/context.test.ts
```diff
@@ -0,0 +1,3068 @@
...
+ */
+
+import anyTest, { TestInterface } from 'ava';
+import * as sinon from 'sinon';
+import { CanvasRenderingContext2DShim, deferredUpgrades } from '../../worker-thread/canvas/CanvasRenderingContext2D';
+import { HTMLCanvasElement } from '../../worker-thread/dom/HTMLCanvasElement';
+import { CanvasRenderingContext2D } from '../../worker-thread/canvas/CanvasTypes';
+import { createTestingDocument } from '../DocumentCreation';
+
+const test = anyTest as TestInterface<{
```
choumx: This test is 3K lines of code. We should first remove redundant tests (no need to test same upgrade code path more than once) and then split it up if necessary (how should we segment it?).

src/test/htmlcanvaselement/context.test.ts
```diff
@@ -0,0 +1,3068 @@
...
+    t.true(stub.withArgs(1, 2, 3, 4).calledOnce);
+    t.false(implStub.called);
+  });
+});
+
+test('context calls both versions of clearRect when called before upgrade', async t => {
+  const { context2d, deferredUpgrade, implementation, sandbox } = t.context;
+  const instance = new FakeOffscreenCanvas();
+
+  const instanceStub = (instance.getContext('2d')['clearRect'] = sandbox.stub());
```
choumx: I don't understand how this local `FakeOffscreenCanvas` instance is at all related to the `<canvas>` element created in `beforeEach`.

src/test/main-thread/offscreen-canvas.test.ts
```diff
@@ -0,0 +1,258 @@
...
+});
+
+// main-thread's strings API does not return an ID when storing a string
+// so for convenience:
+function storeString(strings: Strings, text: string, currentIndex = -1) {
+  strings.store(text);
+  return ++currentIndex;
+}
+
+function exectuteCall(
```
choumx: "exectute"

PR #381, opened 2019-03-22
A babel plugin to provide warnings on uses of sync properties/functions

dvoytenko:
TODO:

- [x] tests
- [x] good name
- [x] docs
- [x] complete the dictionary


tools/tool1/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+            (spec.global
+             && parent.type == 'CallExpression'
+             && parentKey == 'callee')
+            ||
+            // Global property reads.
+            (spec.global
+             && path.isReferencedIdentifier());
+
+        // Max depth is 3 for a function call.
+        const hasOk = hasOkComment(node)
```
jridgewell: This is a weird to do, and it's only necessary because we're handling the three AST nodes at the `Identifier` level. Splitting into explicit node visitors (`ReferencedIdentifier`, `MemberExpression`, and `Identifier`) would help a little. We could even visit `ObjectPattern` instead of `Identifier`.

tools/tool1/src/index.js
```diff
@@ -0,0 +1,90 @@
...
+        }
+        const spec = DICT[name];
+        if (!spec) {
+          return;
+        }
+
+        const filtered =
+            // Member properties and function calls.
+            (parent.type == 'MemberExpression'
+             && parentKey == 'property')
```
jridgewell: Best practice is to use babel-types:

```js
t.isMemberExpression(parent, { key: node } )
```

jridgewell: Additionally, this will fire for computed properties where the property key is an identifier with the same name (but the actual property read by the key could be safe).

dvoytenko: Done. Here and for all other similar.

tools/tool1/src/index.js
dvoytenko: Yeah. I tried that. The problem was that in the end I still had to use `Identifier` for the simplest statement-based expressions anyway. So it kind of ended up doing the same work twice. I know this makes some comment situations a bit odd, but it looked like a cleaner compromise.

tools/tool1/src/index.js
dvoytenko: I believe I fixed that in the last version. PTAL.

tools/tool1/src/index.js
```diff
@@ -0,0 +1,104 @@
...
+
+        const parentOfParentPath = parentPath && parentPath.parentPath;
+        const parentOfParent = parentOfParentPath && parentOfParentPath.node;
+        const parentOfParentType =
+            parentOfParentPath && parentOfParentPath.node.type;
+
+        const filtered =
+            // Member properties and function calls.
+            // `btn.offsetWidth`, `btn.getBoundingClientRect()`
+            (t.isMemberExpression(parent, {property: node}))
```
jridgewell: This will still fire for `btn[offsetWidth]`, where we don't actually know what `offsetWidth` will resolve to the string `"offsetWidth"`.

tools/tool1/src/index.js
```diff
@@ -0,0 +1,104 @@
...
+            ||
+            // Object property pattern in declaration.
+            // `const {offsetWidth} = btn`
+            (t.isObjectProperty(parent, {key: node})
+             && !path.isReferencedIdentifier()
```
jridgewell: This is guaranteed by `t.isObjectProperty(parent, {key: node})`

tools/tool1/src/index.js
dvoytenko: Huh. That's a weird AST resolution, isn't it? Is there any difference between actual property name vs some variable in brackets?

dvoytenko: Looks like I can restrict binding of the `btn[offsetWidth]`, which makes it ok.

tools/tool1/src/index.js
dvoytenko: Done

tools/tool1/src/index.js
dvoytenko: Supported string literals and bound/unbound resolutions as well. PTAL.

tools/tool1/src/index.js
```diff
@@ -0,0 +1,119 @@
...
+  }
+
+  const hasBinding = isLiteral ? false : scope.hasBinding(name);
+  const parentOfParentPath = parentPath && parentPath.parentPath;
+  const parentOfParent = parentOfParentPath && parentOfParentPath.node;
+
+  const filtered =
+      // Member properties and function calls.
+      // `btn.offsetWidth`, `btn.getBoundingClientRect()`
+      // `btn['offsetWidth']`, `btn['getBoundingClientRect']()`
```
jridgewell: What I meant is that you should be checking if the property is computed, if it is, skip it.

tools/tool1/src/index.js
```diff
@@ -0,0 +1,119 @@
...
+ * limitations under the License.
+ */
+
+const DICT = require('./dict');
+
+
+module.exports = function({types: t}) {
+  return {
+    visitor: {
+      Identifier(path) {
```
jridgewell: These are going to be very slow visitors, because there thousands of identifiers/literals in a project.

dvoytenko: Definitely, but I can't avoid scanning for all identifiers, can i? Or is there a clear way for me to pick up all the

```
pageXOffset;
x = pageXOffset;
if (pageXOffset) {}
...
```
?

jridgewell: `ReferencedIdentifier` will fire for those cases.

dvoytenko: Awesome! This is exactly what I was missing. I reimplemented the whole thing as you recommended for separate expression types. PTAL.

tools/tool1/src/index.js
dvoytenko: Done

tools/tool1/src/index.js
dvoytenko: Redid with separate expressions per your recommendation.

tools/tool1/src/index.js
```diff
@@ -0,0 +1,182 @@
...
+        }
+        const { name } = node;
+        if (!name) {
+          return;
+        }
+        const spec = DICT[name];
+        if (!spec) {
+          return;
+        }
+        if (hasOkComment(node, parent, parentPath.parent)) {
```
kristoferbaxter: What does the OK comment do for this tool? Designed to catch false positives?

tools/tool1/src/dict.js
```diff
@@ -0,0 +1,60 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+
+/**
+ * @const {!Object<string, {
```
kristoferbaxter: What's the motivation for using CC style annotations instead of TypeScript like `WorkerDOM` uses?

tools/tool1/src/dict.js
```diff
@@ -0,0 +1,60 @@
...
+ *   global: (boolean|undefined),
+ *   replacement: (string|undefined),
+ * }>}
+ */
+module.exports = {
```
kristoferbaxter: Could this file be converted to JSON and imported? Static defined configuration is quite nice where doable.

tools/tool1/src/index.js
```diff
@@ -0,0 +1,182 @@
...
+        if (!properties || properties.length == 0) {
+          return;
+        }
+
+        for (const property of properties) {
+          const { key } = property;
+          if (!key) {
+            continue;
+          }
+          const name =
```
kristoferbaxter: Is this formatted using the `prettier` configuration for the remainder of the project?

I don't think its running by default on `.js` payloads, would make style mostly irrelevant for review.

tools/tool1/test/fixtures/not-matching.js
```diff
@@ -0,0 +1,37 @@
...
+x = {pageXOffset: 1};
+offsetWidth.other;
+getBoundingClientRect.other();
+btn.other;
+btn['other'];
+btn[1];
+function one(pageXOffset) {
+  pageXOffset;
+}
+function two(getComputedStyle) {
```
kristoferbaxter: When all the cases are bundled together in one file, comments would be useful to describe why this specific code segment shouldn't match.

tools/tool1/src/index.js
dvoytenko: Yes. They are quite possible and no clear way to eliminate them.

tools/tool1/src/dict.js
dvoytenko: This is only for documentation. But do you mean you'd like to port to TypeScript? Or just use TypeScript notations? I pickpocketed pieces of this from other plugins that were all JS-based. Not sure how much type definition is missing to turn this to TypeScript code, but let me know what you'd prefer.

tools/tool1/src/dict.js
dvoytenko: I generally prefer this because there's a clear way to import and, provided compiler is setup this way, this can easily participate in compiler rules/obfuscation/minification/rollup. Plus the syntax is significantly better. What's your reasoning for JSON?

tools/tool1/src/index.js
kristoferbaxter: Makes sense, let's throw a note in the readme when we're closer to complete.

tools/tool1/src/dict.js
kristoferbaxter: Either approach is fine with me.

Why JSON? JSON is static by default and encourages a different authoring mindset (less likely to add executable functions as values to keys, etc).

tools/tool1/src/index.js
kristoferbaxter: #386 Submitted to address this.

dvoytenko: done

tools/tool1/test/fixtures/not-matching.js
dvoytenko: Adding comments explaining each case.

tools/tool1/src/index.js
dvoytenko: Added a note to readme.

tools/tool1/src/dict.js
dvoytenko: I see. I'd still rather keep it as an JS import. I often find JSON barely readable...

tools/tool1/package.json
```diff
@@ -0,0 +1,40 @@
+{
+  "name": "@ampproject/worker-dom-tools-tool1",
```
kristoferbaxter: Like the name `@ampproject/worker-dom-linter`?

tools/tool1/src/index.js
```diff
@@ -0,0 +1,161 @@
...
+
+        for (const property of properties) {
+          const { key } = property;
+          if (!key) {
+            continue;
+          }
+          const name = t.isStringLiteral(key) ? key.value : t.isIdentifier(key) ? key.name : null;
+          if (!name) {
+            continue;
+          }
```
jridgewell: This requires the same `node.computed && t.isIdentifier(key)` as `MemberExpression`

tools/tool1/src/index.js
```diff
@@ -0,0 +1,161 @@
...
+        const { node, parent } = path;
+        if (!node || !parent) {
+          return;
+        }
+        const { properties } = node;
+        if (!properties || properties.length == 0) {
+          return;
+        }
+
+        for (const property of properties) {
```
jridgewell: Nit: use a regular for loop.

tools/tool1/src/index.js
dvoytenko: Added and added a test for this.

tools/tool1/src/index.js
dvoytenko: Done

tools/tool1/src/dict.js
dvoytenko: Redid in TypeScript style.

tools/tool1/package.json
dvoytenko: Done

PR #393, opened 2019-03-27
Boolean properties

kristoferbaxter:
This PR adds support for Boolean properties over the Transferable interface.


src/worker-thread/dom/HTMLOptionElement.ts
```diff
@@ -61,17 +66,23 @@ export class HTMLOptionElement extends HTMLElement {
    * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement
    * @return boolean based on if the option element is selected.
    */
-  get selected(): boolean {
+  get selected(): boolean | string {
```
dvoytenko: I don't quite understand this part. Option's `selected` property is always boolean, no?

src/worker-thread/dom/HTMLOptionElement.ts
```diff
@@ -61,17 +66,23 @@ export class HTMLOptionElement extends HTMLElement {
...
 
   /**
    * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement
    * @param value new selected boolean value.
    */
-  set selected(value: boolean) {
-    this[TransferrableKeys.selected] = value;
-    // TODO(KB) This is a mutation.
+  set selected(value: boolean | string) {
+    this[TransferrableKeys.selected] = typeof value === 'boolean' ? value : value !== '';
```
dvoytenko: Not sure this is right per spec. I think any value can be supplied and it's just evaluated as truthy. No?

src/worker-thread/dom/HTMLOptionElement.ts
kristoferbaxter: In TypeScript the getter and setter must have an equivalent type, even though we always return a boolean.

kristoferbaxter: Changed to any, since the official DOM API accepts any value in the setter.

src/worker-thread/dom/HTMLOptionElement.ts
kristoferbaxter: Absolutely right, moved to just a truthy check.

PR #397, opened 2019-03-28
Enable value setter for HTMLSelectElement

kristoferbaxter:
`HTMLSelectElement.set value` wasn't implemented, and is needed for a demo I am working on.

This PR changes `Node` to allow for a new method that runs after insertion or removal, but before the microtask collection in `MutationObserver`. This lets extensions of `Node` like `HTMLSelectElement` update local information on each insertion/removal.

### DIFF
--- a/src/worker-thread/dom/HTMLSelectElement.ts
+++ b/src/worker-thread/dom/HTMLSelectElement.ts
@@ -21,9 +21,10 @@ import { HTMLInputLabelsMixin } from './HTMLInputLabelsMixin';
 import { matchChildrenElements, matchChildElement, tagNameConditionPredicate } from './matchElements';
 import { HTMLOptionElement } from './HTMLOptionElement';
 import { TransferrableKeys } from '../../transfer/TransferrableKeys';
+import { Node } from './Node';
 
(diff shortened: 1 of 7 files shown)


src/worker-thread/dom/HTMLSelectElement.ts
```diff
@@ -118,10 +151,19 @@ export class HTMLSelectElement extends HTMLElement {
...
     return firstSelectedChild ? (firstSelectedChild as HTMLOptionElement).value : '';
   }
+
+  /**
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement
+   * @set value
+   */
+  set value(value: any) {
+    const stringValue = String(value);
+    this.children.forEach((element: Element) => isOptionPredicate(element) && (element.selected = element.value === stringValue));
```
dvoytenko: Not sure here it's meant to be `===` or `==` per how Web Spec works.

kristoferbaxter: Added some tests here, we always coerce the value to be a string similar to the DOM spec.

PR #418, opened 2019-04-08
Add Support for Event.pageX and Event.pageY

kristoferbaxter:
Quite helpful for mouse events.


kristoferbaxter:
We're missing test coverage for Events dispatched from main thread to the Worker. Need to address soon.

PR #420, opened 2019-04-09
Remove DOMPurify and refactor Sanitizer API

choumx:
- Removes DOMPurify from worker-dom. Instead, require callers to provide an object that implements the `Sanitizer` interface in the `WorkerDOMConfiguration` object.
- Made some Sanitizer API changes to (1) include attribute removals and (2) delegate actual mutation to caller to support value rewriting.


src/main-thread/index.safe.ts
```diff
@@ -14,36 +14,17 @@
...
-
-/** Users can import this and configure the sanitizer with custom DOMPurify hooks, etc. */
-export const sanitizer = new DOMPurifySanitizer();
-
 /**
- * @param baseElement
- * @param domURL
+ * @fileoverview This entry point API in active development and unstable.
  */
-export function upgradeElement(baseElement: Element, domURL: string, longTask?: LongTaskFunction): Promise<Worker | null> {
```
dvoytenko: @kristoferbaxter was explaining to me that there's a key difference with this method where `domURL` actually plays a part. He wanted to remove `domURL` from the config and instead put it here as an argument.

src/worker-thread/index.safe.ts
```diff
@@ -112,7 +112,7 @@ const WHITELISTED_GLOBALS = [
   'encodeURIComponent',
   'escape',
   'eval',
-  'fetch', // TODO(choumx): Wrap this for origin whitelisting.
```
dvoytenko: Are we skipping the "origin whitelisting" work?

src/main-thread/DOMPurifySanitizer.ts
```diff
@@ -1,109 +0,0 @@
...
- * DOMPurify hooks before/after sanitization.
- * @see https://github.com/cure53/DOMPurify#hooks
- */
-interface DOMPurifySanitizerCallbacks {
-  afterSanitize?: (purify: DOMPurify) => void;
-  beforeSanitize?: (purify: DOMPurify) => void;
-  nodeWasRemoved?: (node: Node) => void;
-}
-
-export class DOMPurifySanitizer implements Sanitizer {
```
dvoytenko: This is an interesting situation. Can we actually setup an interface implementation over the boundary between AMP's Closure codebase and WorkerDOM's TypeScript? Can we even export `Sanitizer` interface over to Closure?

choumx: I don't think so, unless we export a types declaration from worker-dom and convert amp-script to TS. 

I actually added support for extensions in TS and transpiled to Closure-annotated JS via tsickle awhile ago. I'll file an issue for this.

dvoytenko: Ok. Should I rereview? Or not yet?

src/worker-thread/index.safe.ts
choumx: Update: Tested this and looks like enforcement via `<meta>` CSP is possible. Will follow up with this separately.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy#Multiple_content_security_policies

PR #427, opened 2019-04-10
Touch Data on Events

kristoferbaxter:
We're lacking tests in this area of the codebase, but we now have support for touch event data in the Worker thread.


src/main-thread/commands/event-subscription.ts
```diff
@@ -105,6 +126,8 @@ export function EventSubscriptionProcessor(strings: Strings, nodeContext: NodeCo
         [TransferrableKeys.keyCode]: 'keyCode' in event ? event.keyCode : undefined,
         [TransferrableKeys.pageX]: 'pageX' in event ? event.pageX : undefined,
         [TransferrableKeys.pageY]: 'pageY' in event ? event.pageY : undefined,
+        [TransferrableKeys.touches]: createTransferrableTouch(event, 'touches'),
+        [TransferrableKeys.changedTouches]: createTransferrableTouch(event, 'changedTouches'),
```
choumx: Nit: Would be a bit more consistent/readable if the `undefined` condition was extracted here:

```ts
// Counterpart is touchListFromTransfer().
function touchListToTransfer(touchList: TouchList): TransferrableTouchList {
  return Object.values(touchList).map(touch => {
    return [touch.identifier, touch.screenX, ...];
  });
}

[TransferrableKeys.touches]: 'touches' in event ? touchListToTransfer(event.touches) : undefined,
[TransferrableKeys.touches]: 'changedTouches' in event ? touchListToTransfer(event.changedTouches) : undefined,
```

src/transfer/TransferrableEvent.ts
```diff
@@ -17,6 +17,21 @@
...
+interface TransferrableTouch {
+  [key: number]: number;
+  // identifier
+  // screenX
+  // screenY
+  // clientX
+  // clientY
+  // pageX
+  // pageY
+  // target._index_
```
choumx: Why are these fields commented out? Shouldn't type checker complain about this?

src/worker-thread/Event.ts
```diff
@@ -80,6 +97,40 @@ const targetFromTransfer = (document: Document, event: TransferrableEvent): Node
...
+        pageX: transferredTouch[5],
+        pageY: transferredTouch[6],
+        target: get(transferredTouch[7] !== 0 ? transferredTouch[7] : document[TransferrableKeys.index]),
+      };
+    });
+
+    return list;
+  }
+  return undefined;
+};
```
choumx: Ditto for this function and callers below.

src/transfer/TransferrableEvent.ts
```diff
@@ -17,6 +17,21 @@
...
+}
+export interface TransferrableTouchList {
+  [key: number]: TransferrableTouch;
```
choumx: Shouldn't interface contain an array of `TransferrableTouch`?

src/transfer/TransferrableEvent.ts
kristoferbaxter: I'll add a comment, but this is similar to the other Transferables where the index indicates a specific item, and a const enum lets you access specific members in the correct locations without referring to this comment each time.

src/transfer/TransferrableEvent.ts
kristoferbaxter: We can change the interface here, but these are functionally equivalent for typing.

src/main-thread/commands/event-subscription.ts
```diff
@@ -65,6 +65,27 @@ const fireResizeChange = (workerContext: WorkerContext, cachedWindowSize: [numbe
...
+        touch.identifier,
+        touch.screenX,
+        touch.screenY,
+        touch.clientX,
+        touch.clientY,
+        touch.pageX,
+        touch.pageY,
+        (touch.target as RenderableElement)._index_,
+      ])
+    : undefined;
```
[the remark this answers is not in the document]
kristoferbaxter: Attempting to be consistent with the other methods defined in this module. Perhaps we should tackle them all together in a separate PR?

src/main-thread/commands/event-subscription.ts
kristoferbaxter: That's far more consistent with the pattern here. Good call, will change!

src/main-thread/commands/event-subscription.ts
```diff
@@ -65,6 +65,27 @@ const fireResizeChange = (workerContext: WorkerContext, cachedWindowSize: [numbe
     [TransferrableKeys.sync]: cachedWindowSize,
   });
 
+/**
+ * Convert a TouchList into a TransferrableTouchList
+ * @param touches
+ */
+const createTransferrableTouch = (
+  event: Event | KeyboardEvent | MouseEvent | TouchEvent,
```
dvoytenko: A bit lost: why is `KeyboardEvent` here?

src/worker-thread/Event.ts
```diff
@@ -30,6 +30,21 @@ interface EventOptions {
...
+  readonly screenY: number;
+  readonly clientX: number;
+  readonly clientY: number;
+  readonly pageX: number;
+  readonly pageY: number;
+  readonly target: Node | null;
+}
+interface TouchList {
+  [key: number]: Touch;
+  length: number;
```
dvoytenko: Isn't this one of those APIs that also have an `item()` method? I think they are still used quite a bit, despite the `[i]` accessor.

src/main-thread/commands/event-subscription.ts
kristoferbaxter: Fixed. Thanks!

src/worker-thread/Event.ts
kristoferbaxter: Added `item(index: number) => Touch | null` method.

PR #466, opened 2019-05-03
Move to Xenial

kristoferbaxter:
Trying out Xenial instead of Trusty.


kristoferbaxter:
Looks like this makes the build take about 20% longer on Linux.

PR #467, opened 2019-05-04
Remove Whitespace from Worker Template Literal 

kristoferbaxter:
Worker source template literal contained newlines and extraneous spaces.

This is not an attempt to use a minifier on its source, but rather a small number of safe reductions in character count.

### DIFF
--- a/config/rollup.plugins.js
+++ b/config/rollup.plugins.js
@@ -40,6 +40,9 @@ export function babelPlugin({ transpileToES5, allowConsole = false, allowPostMes
           targets,
           loose: transpileToES5 ? false : true,
           modules: false,
+          // Excluding 'transform-template-literals'... this library doesn't need caching for template literals.
+          // As a result, the Safari bug causing transforms here isn't material.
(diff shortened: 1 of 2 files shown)


config/rollup.plugins.js
```diff
@@ -126,3 +126,32 @@ export function removeDebugCommandExecutors() {
...
+      walk.simple(program, {
+        TemplateLiteral(node) {
+          let literalValue = code.substring(node.range[0], node.range[1]);
+          literalValue = literalValue
+            .replace(/\) \{/g, '){')
+            .replace(/, /g, ',')
+            .replace(/ = /g, '=')
+            .replace(/\t/g, '')
+            .replace(/[ ]{2,}/g, '')
+            .replace(/\n/g, '');
```
patrickkettner: for my own edification, how hot is this loop? I assume the regex is being optimized for readability over perf?

kristoferbaxter: This is run 4 times during the build process, during CI and locally if the commit hook is enabled.

Went for readability since this is an odd scenario, minifying output of a template literal that constructs JavaScript.

PR #473, opened 2019-05-08
Use incremental compilation

jridgewell:
Should help speed up repeated builds of the project. On my 2015 Macbook Pro, this reduced test compilation from 23s to 10s.


package.json
```diff
@@ -23,9 +23,9 @@
     "tsc:build:worker": "tsc -p config/tsconfig.build.worker-thread.json",
     "tsc:build:main": "tsc -p config/tsconfig.build.main-thread.json",
     "clean": "rimraf output dist",
-    "pretest": "npm-run-all clean --parallel tsc:test:*",
+    "pretest": "npm-run-all --parallel tsc:test:*",
     "test": "ava -v output/test/**/*.test.js",
-    "pre~rollup": "npm-run-all clean --parallel tsc:build:*",
```
kristoferbaxter: Like this idea, but we should ensure that the `dist` directory is cleared out before building. Perhaps we should have a `rimraf dist` in a `prebuild` script.

Also, this change removes all usages of the `clean` script, so we should likely remove it.

PR #481, opened 2019-05-13
Minor Code Reductions

kristoferbaxter:
Uses the same patterns for applying mutations in the Canvas Polyfill.


src/transfer/TransferrableKeys.ts
```diff
@@ -82,4 +82,5 @@ export const enum TransferrableKeys {
   hydrateNode = 64,
   offsetX = 65,
   offsetY = 66,
+  mutated = 67,
```
jridgewell: We should probably stop calling these `TransferrableKeys` at some point.

kristoferbaxter: Agreed.

PR #482, opened 2019-05-13
Canvas minify

kristoferbaxter:


src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -37,9 +25,9 @@ import { HTMLElement } from '../dom/HTMLElement';
 
 export const deferredUpgrades = new WeakMap();
 
-export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> implements CanvasRenderingContext2D {
+export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> {
```
jridgewell: Why change this?

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -37,9 +25,9 @@ import { HTMLElement } from '../dom/HTMLElement';
...
   private queue = [] as { fnName: string; args: any[]; isSetter: boolean }[];
-  private implementation: CanvasRenderingContext2D;
+  public [TransferrableKeys.implementation]: CanvasRenderingContext2D;
```
jridgewell: Why make this public?

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -90,33 +77,29 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
+    const returnValue = (this[TransferrableKeys.implementation] as any)[name](...args);
     if (!this.upgraded) {
       this.queue.push({ fnName: name, args, isSetter: false });
     }
     return returnValue;
   }
 
-  private delegateSetter(name: string, args: any[]) {
-    (this.implementation as any)[name] = args[0];
+  public [TransferrableKeys.set](name: string, args: any[]) {
```
jridgewell: Ditto visibility.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -458,13 +285,38 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
+  'shadowOffsetX',
+  'shadowOffsetY',
+  'globalAlpha',
+  'globalCompositeOperation',
+  // Experimental
+  'imageSmoothingEnabled',
+  'imageSmoothingQuality',
+  'filter',
+].forEach(property => {
+  Object.defineProperty(CanvasRenderingContext2DShim.prototype, property, {
```
jridgewell: We discussed this originally, and decided to leave the methods in place to give typescript the type hints. If we don't imagine many people will use the types, we can further meta program this file (and the others from the PR).

src/worker-thread/canvas/CanvasRenderingContext2D.ts
kristoferbaxter: Since the base class no longer implements the getter/setter pairs (it must be augmented by `Object.define...`) the implementation is no longer true during this definition.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
kristoferbaxter: This way `Object.defineProperties` could leverage the implementation. Technically the output doesn't require this since its not using private fields, but it does violate the principle to leverage it outside of the class definition or an extension.

Not great, but I haven't found a better pattern for this.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
kristoferbaxter: I expect that authors will end up using the definitions we publish, not the source files as their type hints.

PR #483, opened 2019-05-13
Incremental TypeScript

kristoferbaxter:
This draft show the change to ensure when creating a release we don't accidentally use the test format of the TypeScript output.

Do you think this change could be made in #473?


jridgewell:
Let's merge this one, and close #473.

PR #490, opened 2019-05-16
Canvas Refactoring / Documenting

delima02:
- Refactor `offscreen-canvas-polyfill` processor tests and `CanvasRenderingContext2DShim` tests
- Add JSDoc comments to all Canvas related classes and some other functions.
- Add comments in tests when the flow is not too obvious
- Remove static property from the `FakeOffscreenCanvas` class that is used for testing

@jridgewell @choumx


src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -67,7 +66,7 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
     // 2. Retrieve an auto-synchronized OffscreenCanvas from the main-thread and call all methods
     // in the queue.
     else {
-      this.implementation = this.unUpgradedOffscreenCanvasContext = new OffscreenCanvas(0, 0).getContext('2d');
+      this.implementation = new OffscreenCanvas(0, 0).getContext('2d');
       this.goodOffscreenPromise = this.getOffscreenCanvasAsync(this.canvasElement).then(instance => {
```
jridgewell: Good work removing the `unUpgradedOffscreenCanvasContext`. The next one we can target is this `goodOffscreenPromise`. There's an approach we can take that will let us remove the property entirely.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -37,34 +37,46 @@ import { HTMLElement } from '../dom/HTMLElement';
...
 
-  private getOffscreenCanvasAsync(canvas: ElementType): Promise<{ getContext(c: '2d'): CanvasRenderingContext2D }> {
-    return new Promise((resolve, reject) => {
+  /**
+   * Retrieves auto-synchronized version of an OffscreenCanvas from the main-thread.
+   * @param canvas HTMLCanvasElement associated with this context.
+   */
+  private getOffscreenCanvasAsync(canvas: ElementType): Promise<void> {
+    let deferred: { resolve?: (value?: {} | PromiseLike<{}> | undefined) => void; upgradePromise?: Promise<void> } = {};
+    let testMode = false; // whether in testing environment
```
jridgewell: Set `testMode` here.

delima02: Like this?

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -37,34 +37,46 @@ import { HTMLElement } from '../dom/HTMLElement';
...
 
-  private getOffscreenCanvasAsync(canvas: ElementType): Promise<{ getContext(c: '2d'): CanvasRenderingContext2D }> {
-    return new Promise((resolve, reject) => {
+  /**
+   * Retrieves auto-synchronized version of an OffscreenCanvas from the main-thread.
+   * @param canvas HTMLCanvasElement associated with this context.
+   */
+  private getOffscreenCanvasAsync(canvas: ElementType): Promise<void> {
+    let deferred: { resolve?: (value?: {} | PromiseLike<{}> | undefined) => void; upgradePromise?: Promise<void> } = {};
+    let testMode = typeof addEventListener !== 'function'; // whether in testing environment
```
choumx: Can we use an environment or bundle-time variable instead of inferring this? Ideally this would compile to `const testMode = false;` in dist/ output. E.g. see `DEBUG_ENABLED`.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -37,34 +37,46 @@ import { HTMLElement } from '../dom/HTMLElement';
...
   }
 
-  private getOffscreenCanvasAsync(canvas: ElementType): Promise<{ getContext(c: '2d'): CanvasRenderingContext2D }> {
-    return new Promise((resolve, reject) => {
+  /**
+   * Retrieves auto-synchronized version of an OffscreenCanvas from the main-thread.
+   * @param canvas HTMLCanvasElement associated with this context.
+   */
+  private getOffscreenCanvasAsync(canvas: ElementType): Promise<void> {
+    let deferred: { resolve?: (value?: {} | PromiseLike<{}> | undefined) => void; upgradePromise?: Promise<void> } = {};
```
choumx: `const` for vars that don't change. Also what's up with the type definition for `resolve`?

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -37,34 +37,46 @@ import { HTMLElement } from '../dom/HTMLElement';
 
 export const deferredUpgrades = new WeakMap();
```
choumx: Why isn't this an instance variable?

src/worker-thread/canvas/CanvasRenderingContext2D.ts
delima02: I wrote this sticking to what TypeScript indicated to be `resolve`'s type. Is something wrong with it?

src/worker-thread/canvas/CanvasRenderingContext2D.ts
choumx: `DEBUG_BUNDLE` is a better example since we don't use rollup in the test pipeline. See package.json.

delima02: Only in testing environment I saw this compile to true. `DEBUG_ENABLED`'s module is not visible to the test suite so it doesn't really help here (we would need a different mechanism).

choumx:
Also, I noticed that we don't set OffscreenCanvas in `GlobalScope` in `src/worker-thread/index(.amp).ts`, so I think native OffscreenCanvas will never be used.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
delima02: Isn't that worse? It would have to be public and only exist in our version of `CanvasRenderingContext2D` (which would be slightly inconsistent).

src/test/DocumentCreation.ts
```diff
@@ -121,10 +123,15 @@ const GlobalScope: GlobalScope = {
...
 };
 
-export function createTestingDocument(additional: {} | null = null): Document {
-  const customGlobal = Object.assign({}, GlobalScope, additional);
+/**
+ * Creates a Document object for testing environment.
+ * @param overrides Can add a new variable to Global or override an existing one.
+ */
+export function createTestingDocument(overrides: {} | null = null): Document {
+  const customGlobal = Object.assign({}, GlobalScope, overrides, { TEST_ENABLED: true });
```
choumx: Ironically, I think we actually want to use a global in this case. Can you think of reasons why this is OK for flags but not for objects under test?

Maybe this will work -- define `(global as any).DEBUG_ENABLED = true;` in this file and replace `TEST_ENABLED` with `DEBUG_ENABLED` in the canvas code.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -76,15 +88,28 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
         addEventListener('message', messageHandler);
         transfer(canvas.ownerDocument as Document, [TransferrableMutationType.OFFSCREEN_CANVAS_INSTANCE, canvas[TransferrableKeys.index]]);
       }
+    }).then((instance: { getContext(c: '2d'): CanvasRenderingContext2D }) => {
+      this.implementation = instance.getContext('2d');
+      this.upgraded = true;
+      this.flushQueue();
     });
+
+    if (TEST_ENABLED) {
```
kristoferbaxter: Values like this should be set and controlled by Rollup replacements.

delima02: Tests don't use Rollup, which means if we do that, the variable will appear as undefined during test execution. This will be deferred to a later PR.

src/worker-thread/index.amp.ts
```diff
@@ -131,6 +131,8 @@ const WHITELISTED_GLOBALS = [
   'unescape',
 ];
 
+declare var OffscreenCanvas: any;
```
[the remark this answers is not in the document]
delima02: I have swapped this for a mock interface. However, it will not build if the type declaration is included in a declaration file (any .d file). It seems rollup can't resolve those. I'll address this in a separate PR.

src/worker-thread/WorkerDOMGlobalScope.ts
```diff
@@ -90,6 +90,7 @@ export interface GlobalScope {
   HTMLTableRowElement: typeof HTMLTableRowElement;
   HTMLTableSectionElement: typeof HTMLTableSectionElement;
   HTMLTimeElement: typeof HTMLTimeElement;
+  OffscreenCanvas: any | undefined;
```
choumx: `typeof OffscreenCanvas`, and import `OffscreenCanvas` from CanvasTypes.d.ts.

PR #549, opened 2019-06-24
CanvasPattern and drawImage Support

delima02:
- Support for `CanvasRenderingContext2D.createPattern`.
- `CanvasPattern` class is used by the polyfill case (similar to the `CanvasGradient` implementation).
- `FakeNativeCanvasPattern` wraps the necessary API for the native case. 
- Added a `CanvasPattern` in the Canvas demo page.
- Implemented `CanvasRenderingContext2D.drawImage`. 
- Since `serializeAsTransferrableObject` method was required on `HTMLElement`, changed the method's name to use a more obscure TransferrableKey. This required some downstream changes.

to/ @jridgewell @choumx @kristoferbaxter


demo/canvas/index.html
```diff
@@ -21,6 +21,7 @@
     </div>
     <div>
       <canvas width="500" height="500" id="myCanvas"></canvas>
+      <img id="myImg" hidden src="https://mdn.mozillademos.org/files/222/Canvas_createpattern.png">
```
kristoferbaxter: Can you make this image point to something loaded from the demo directory?

src/worker-thread/canvas/CanvasPattern.ts
```diff
@@ -0,0 +1,64 @@
...
+ * Wrapper class for CanvasPattern. The user will be able to manipulate as a regular CanvasPattern object.
+ * This class will be used when the CanvasRenderingContext is using an OffscreenCanvas polyfill.
+ */
+export class CanvasPattern implements TransferrableObject {
+  private id: number;
+  private document: Document;
+
+  constructor(id: number, document: Document, args: any[], serializedCreationObject: number[]) {
+    this.id = id;
+    this.document = document;
```
jridgewell: I don't think it's necessary to store the document as a property.

src/worker-thread/canvas/CanvasPattern.ts
```diff
@@ -0,0 +1,64 @@
...
+    this.createObjectReference(serializedCreationObject, args);
+  }
+
+  /**
+   * Creates CanvasPattern object in the main thread, and associates it with this object with the id provided.
+   * @param serializedCreationObject The target object needed to create the corresponding object in the main thread, serialized.
+   * @param creationMethod Method to use to create this object.
+   * @param args Arguments needed for object creation.
+   */
+  private createObjectReference(serializedCreationObject: number[], args: any[]) {
```
jridgewell: I don't think this belongs on the `CanvasPattern`, and it becomes more obvious when we're storing `createPattern` as the method name. This is a `CanvasRenderingContext2D` concern.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -115,9 +125,9 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
   private flushQueue() {
     for (const call of this.queue) {
       if (call.isSetter) {
-        (this.implementation as any)[call.fnName] = call.args[0];
+        (this as any)[call.fnName] = call.args[0];
```
jridgewell: We should be able to avoid this typecast using a [`keyof`](https://www.typescriptlang.org/docs/handbook/advanced-types.html#index-types) type:

```js
let personProps: keyof Person
```

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -252,15 +262,25 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
 
   /* FILL AND STROKE STYLES */
   set fillStyle(value: string | CanvasGradient | CanvasPattern) {
-    this.delegateSetter('fillStyle', [...arguments]);
+    const isGradient = typeof (value as any).addColorStop !== 'undefined';
+    if (typeof value === 'string' || isGradient || this.polyfillUsed || !this.upgraded) {
```
jridgewell: There are 12 possible outcomes in this single test. Please add comments why each one is checked.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -252,15 +262,25 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
   }
 
   get fillStyle(): string | CanvasGradient | CanvasPattern {
     return this.delegateGetter('fillStyle');
   }
 
   set strokeStyle(value: string | CanvasGradient | CanvasPattern) {
-    this.delegateSetter('strokeStyle', [...arguments]);
+    const isGradient = typeof (value as any).addColorStop !== 'undefined';
+    if (typeof value === 'string' || isGradient || this.polyfillUsed || !this.upgraded) {
```
jridgewell: Ditto.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -252,15 +262,25 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
+    }
   }
 
   get fillStyle(): string | CanvasGradient | CanvasPattern {
     return this.delegateGetter('fillStyle');
   }
 
   set strokeStyle(value: string | CanvasGradient | CanvasPattern) {
-    this.delegateSetter('strokeStyle', [...arguments]);
+    const isGradient = typeof (value as any).addColorStop !== 'undefined';
```
jridgewell: I think `instanceof` checks would be more clear here.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -277,7 +297,86 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
   }
 
   createPattern(image: CanvasImageSource, repetition: string): CanvasPattern | null {
-    return this.delegateFunc('createPattern', [...arguments]);
+    const isBitmap = typeof (image as any)[TransferrableKeys.serializeAsTransferrableObject] === 'undefined';
```
jridgewell: Ditto `instanceof`.

src/worker-thread/canvas/FakeNativeCanvasPattern.ts
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  /**
+   * Retrieves ImageBitmap object from main-thread, which replicates the input image. Upon
+   * retrieval, uses it to create a real CanvasPattern and upgrade the implementation property.
+   * @param canvas Canvas element used to create the pattern.
+   * @param image Image to be used as the pattern's image
+   * @param repetition DOMStrings indicating how to repeat the pattern's image.
+   */
+  getCanvasPatternAsync(canvas: ElementType, image: CanvasImageSource, repetition: string): Promise<void> {
```
jridgewell: This is exposed as a public method on an instance the user will receive.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -277,7 +297,86 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
+      const OffscreenCanvas = this.canvasElement.ownerDocument.defaultView.OffscreenCanvas;
+      this.implementation = new OffscreenCanvas(0, 0).getContext('2d');
+      this.unresolvedCalls++;
+
+      const fakePattern = new FakeNativeCanvasPattern<ElementType>();
+      fakePattern.getCanvasPatternAsync(this.canvas, image, repetition).then(() => {
+        this.unresolvedCalls--;
+
+        // The good implementation must only come after the last unresolved createPattern call
+        if (this.unresolvedCalls === 0) {
```
jridgewell: There's still a race condition (not here, but I don't want to comment where it actually is):

```js
// In worker thread:
const canvas = document.createElement('canvas');
document.body.appendChild(canvas);

const ctx = cavas.getContext('2d');
const pattern = ctx.createPattern(img);

ctx.strokeStyle = pattern;
ctx.lineWidth = 5
ctx.beginPath();
ctx.strokeRect(0, 0, 25, 25);
ctx.closePath();
```

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -277,7 +297,86 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
+            data[TransferrableKeys.target][0] === (image as any)[TransferrableKeys.index]
+          ) {
+            removeEventListener('message', messageHandler);
+            const transferredImageBitmap = (data as ImageBitmapToWorker)[TransferrableKeys.data];
+            resolve(transferredImageBitmap);
+          }
+        };
+
+        if (typeof addEventListener !== 'function') {
+          throw new Error('addEventListener not a function!');
```
jridgewell: Doesn't this happen in tests?

src/worker-thread/canvas/CanvasRenderingContext2D.ts
delima02: Per our discussion: the difference between setters and functions makes using `keyof` overly complicated to convince typescript we're doing the right thing. Therefore, leave as is.

src/worker-thread/canvas/canvas-utils.ts
```diff
@@ -0,0 +1,45 @@
...
+import { HTMLCanvasElement } from '../dom/HTMLCanvasElement';
+import { Document } from '../dom/Document';
+import { transfer } from '../MutationTransfer';
+
+export function retrieveImageBitmap(image: HTMLImageElement | HTMLCanvasElement, canvas: HTMLCanvasElement): Promise<any> {
+  return new Promise(resolve => {
+    const messageHandler = ({ data }: { data: ImageBitmapToWorker }) => {
+      if (
+        data[TransferrableKeys.type] === MessageType.IMAGE_BITMAP_INSTANCE &&
+        data[TransferrableKeys.target][0] === (image as any)[TransferrableKeys.index]
```
jridgewell: Is this typecast necessary?

src/worker-thread/canvas/canvas-utils.ts
```diff
@@ -0,0 +1,45 @@
...
+import { MessageType, ImageBitmapToWorker } from '../../transfer/Messages';
+import { TransferrableMutationType } from '../../transfer/TransferrableMutation';
+import { HTMLImageElement } from '../dom/HTMLImageElement';
+import { HTMLCanvasElement } from '../dom/HTMLCanvasElement';
+import { Document } from '../dom/Document';
+import { transfer } from '../MutationTransfer';
+
+export function retrieveImageBitmap(image: HTMLImageElement | HTMLCanvasElement, canvas: HTMLCanvasElement): Promise<any> {
+  return new Promise(resolve => {
+    const messageHandler = ({ data }: { data: ImageBitmapToWorker }) => {
```
jridgewell: This is prone to a race condition.

```js
const img = document.querySelector('img');
ctx.drawImage(img);

img.src = 'foobar.com/img.jpg';
ctx.drawImage(img);
```

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -100,8 +112,13 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
       }
     }).then((instance: { getContext(c: '2d'): CanvasRenderingContext2D }) => {
       this.implementation = instance.getContext('2d');
-      this.upgraded = true;
-      this.flushQueue();
+      this.goodImplementation = this.implementation;
```
jridgewell: We still have a race condition here.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -252,15 +269,31 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
-    this.delegateSetter('fillStyle', [...arguments]);
+    // 1. Native pattern instances given to the user hold the 'real' pattern as their implementation prop.
+    // 2. Pattern must be upgraded, otherwise an undefined 'implementation' will be queued instead of the wrapper object.
+    if (value instanceof FakeNativeCanvasPattern && this.upgraded) {
+      this.delegateSetter('fillStyle', [(value as any).implementation]);
+
+      // Any other case does not require special handling.
+    } else {
+      this.delegateSetter('fillStyle', [...arguments]);
+    }
```
jridgewell: I thought of a new race condition here:

```js
const pattern = ctx.createPattern(img);

ctx2.fillStyle = pattern;
```

Notice I'm passing the pattern to a different canvas' context.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -277,7 +310,69 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
+
+    // Only HTMLElement image sources require special handling. ImageBitmap is OK to use.
+    if (this.polyfillUsed || image instanceof ImageBitmap) {
+      this.delegateFunc('drawImage', [...arguments]);
+    } else {
+      // Queue the drawImage call to make sure it gets called in correct order
+      const args = [] as any[];
+      this.queue.push({ fnName: 'drawImage', args, isSetter: false });
+
+      // Degrade the underlying implementation because we don't want calls on the real one
```
jridgewell: This is still duplicated between `createPattern` and `drawImage`.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -277,7 +310,69 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
+      this.unresolvedCalls++;
+
+      // Retrieve an ImageBitmap from the main-thread with the same image as the input image
+      retrieveImageBitmap(image as any, (this.canvas as unknown) as HTMLCanvasElement)
+        // Then call the actual method with the retrieved ImageBitmap
+        .then((instance: ImageBitmap) => {
+          args.push(instance, dx, dy);
+          this.unresolvedCalls--;
+
+          if (this.unresolvedCalls === 0) {
```
jridgewell: Ditto.

src/worker-thread/canvas/FakeNativeCanvasPattern.ts
```diff
@@ -39,34 +37,20 @@ export class FakeNativeCanvasPattern<ElementType extends HTMLElement> {
...
-    }).then((instance: ImageBitmap) => {
-      const pattern = canvas.getContext('2d').createPattern(instance, repetition);
-
-      if (!pattern) {
-        throw new Error('Pattern is null!');
-      }
-
-      this.realPattern = pattern;
-    });
+  [TransferrableKeys.retrieveCanvasPattern](canvas: ElementType, image: CanvasImageSource, repetition: string): Promise<void> {
```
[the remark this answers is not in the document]
delima02: The `canvas` property in `CanvasRenderingContext2D` is of ElementType, which I can't cast to `HTMLCanvasElement`. But yes, this is intended to always be a Canvas element.

src/worker-thread/canvas/canvas-utils.ts
delima02: Not sure why I added that there. Good catch.

src/main-thread/commands/image-bitmap.ts
```diff
@@ -0,0 +1,55 @@
...
+import { TransferrableMutationType, ImageBitmapMutationIndex } from '../../transfer/TransferrableMutation';
+import { TransferrableKeys } from '../../transfer/TransferrableKeys';
+import { MessageType } from '../../transfer/Messages';
+
+export const ImageBitmapProcessor: CommandExecutorInterface = (strings, nodeContext, workerContext, objectContext, config) => {
+  const allowedExecution = config.executorsAllowed.includes(TransferrableMutationType.IMAGE_BITMAP_INSTANCE);
+
+  return {
+    execute(mutations: Uint16Array, startPosition: number, target: RenderableElement): number {
+      const callIndex = mutations[startPosition + ImageBitmapMutationIndex.CallIndex];
```
jridgewell: Nit: We can inline this into the message back.

src/main-thread/commands/image-bitmap.ts
```diff
@@ -0,0 +1,55 @@
...
+          console.error(`getNode() yields null - ${target}`);
+        }
+      }
+
+      return startPosition + ImageBitmapMutationIndex.End;
+    },
+    print(mutations: Uint16Array, startPosition: number, target?: RenderableElement | null): Object {
+      return {
+        type: 'IMAGE_BITMAP_INSTANCE',
+        target,
```
jridgewell: Need callindex.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -252,15 +285,52 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
 
   set strokeStyle(value: string | CanvasGradient | CanvasPattern) {
-    this.delegateSetter('strokeStyle', [...arguments]);
+    // 1. Native pattern instances given to the user hold the 'real' pattern as their implementation prop.
+    // 2. Pattern must be upgraded, otherwise an undefined 'implementation' could be queued instead of the wrapper object.
+    if (value instanceof FakeNativeCanvasPattern && this.upgraded) {
+      // This case occurs only when an un-upgraded pattern is passed into a different (already
+      // upgraded) canvas context.
+      if (!value[TransferrableKeys.patternUpgraded]) {
+        this.queue.push({ fnName: 'strokeStyle', args: [value], isSetter: true });
```
jridgewell: This seems like the non-pattern path, no?

PR #569, opened 2019-07-10
Fix empty attribute value being processed as removals

### LINKED ISSUE #562: Bug: `el.setAttribute('name', '')`
## Description

```
el.setAttribute('some-attr-name', '')
```

The actual value received in main thread  is `null`. 

This does not match the real DOM world's rule.

## Reproduction

[jsbin](https://jsbin.com/kowexoqoyi/1/edit?html,output)

choumx:
Fixes #562. Good old truthy footgun.

/to @kristoferbaxter


choumx:
@kristoferbaxter Is there a test file for `AttributeProcessor`? I poked around and couldn't find one.

PR #570, opened 2019-07-10
Add "amp/worker/worker.js" output

choumx:
With debugging/console disabled, transpiled to ES5, and compiled/tersed.


kristoferbaxter:
AMP is using both the module build for development, and the ES5 transpiled build for production. It makes sense to publish both for this case.

PR #571, opened 2019-07-11
Allowlist indexedDB, Strings->StringContext, other cleanup

choumx:
Also rename `Strings` to `StringContext` to be consistent with other "Context" classes, and other minor clean up items.

/to @kristoferbaxter

### DIFF
--- a/src/test/main-thread/deserializeTransferrableObject.test.ts
+++ b/src/test/main-thread/deserializeTransferrableObject.test.ts
@@ -17,13 +17,13 @@
 import anyTest, { TestInterface } from 'ava';
 import { TransferrableObjectType } from '../../transfer/TransferrableMutation';
 import { deserializeTransferrableObject } from '../../main-thread/deserializeTransferrableObject';
-import { Strings } from '../../main-thread/strings';
+import { StringContext } from '../../main-thread/strings';
 import { NodeContext } from '../../main-thread/nodes';
 import { Env } from './helpers/env';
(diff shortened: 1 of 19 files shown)


src/test/main-thread/deserializeTransferrableObject.test.ts
```diff
@@ -33,19 +33,19 @@ test.beforeEach(t => {
...
 
-  const strings = new Strings();
+  const stringContext = new StringContext();
   const objectContext = new ObjectContext();
-  const nodeContext = new NodeContext(strings, baseElement);
+  const nodeContext = new NodeContext(stringContext, baseElement);
 
   t.context = {
-    strings,
+    stringContext: stringContext,
```
kristoferbaxter: Nit: no need to define the key and value for equivalent names.

`{ stringContext }` is equivalent to `{ stringContext: stringContext }`

src/test/main-thread/deserializeTransferrableObject.test.ts
```diff
@@ -33,19 +33,19 @@ test.beforeEach(t => {
...
     nodeContext,
     objectContext,
   };
 });
 
 test('Deserializes int arguments', t => {
-  const { strings, nodeContext } = t.context;
+  const { stringContext: strings, nodeContext } = t.context;
```
kristoferbaxter: Nit: no need to map `stringContext` to the local `strings`, since the usage can be fixed on line 52.

PR #573, opened 2019-07-11
Add localStorage/sessionStorage support

### LINKED ISSUE #551: Support localStorage
## Description
According to [WorkerDOM compatibility table](https://github.com/ampproject/worker-dom/blob/master/web_compat_table.md), `Window.localStorage` has been supported. 

> 

Window.localStorage | ✔️
-- | --

However, when I try accessing `localStorage`'s APIs in the worker thread, all of them are `undefined`. And the value of `WorkerThread.workerDom.localStorage` is just a plain object.
## Expected Result
All of the APIs of `localStorage` should exist and return correct results in worker thread. 
## Actual Result
In the scope of author script, `localStorage` is just a plain object and all of its APIs can not be accessed.

choumx:
Fixes #551. ~Includes changes from #571 and #572 (merge those first).~

Hooks up `localStorage` and `sessionStorage`. Allows delegation of storage retrieval and update to the `Sanitizer` (if present). 

Maybe we should rename "Sanitizer" to something more general purpose now?


src/main-thread/commands/storage.ts
```diff
@@ -0,0 +1,86 @@
...
+        const value = valueIndex > 0 ? strings.get(valueIndex) : null;
+
+        // TODO(choumx): Clean up key/value strings (or don't store them in the first place)
+        // to avoid leaking memory.
+
+        if (config.sanitizer) {
+          config.sanitizer.changeStorage(scope, key, value);
+        } else {
+          let storage;
+          if (scope == 'local') {
```
kristoferbaxter: If `scope` is an enum, let's use a const enum instead of the string pool.

In `src/transfer`, there are modules for different types of Transferrable Objects. This is a new type.

```typescript
export const enum StorageLocation {
  local = 0,
  session = 1,
}
```

This would change the `scope` definition above too.
```typescript
const scope = mutations[startPosition + StorageMutationIndex.Scope];
```

And these condition checks would change.
```typescript
if (scope === StorageLocation.local) {
  storage = window.localStorage;
} else if (scope === StorageLocation.session) {
  storage = window.sessionStorage;
}
```

src/main-thread/commands/storage.ts
```diff
@@ -0,0 +1,86 @@
...
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CommandExecutorInterface } from './interface';
+import { TransferrableMutationType, StorageMutationIndex } from '../../transfer/TransferrableMutation';
+
+const CONTEXT = 'STORAGE';
```
kristoferbaxter: An item to remember when building `CommandExecutor`s. The `print` method is only available when debugging the library, as a result duplication doesn't occur in the released copies.

Here `CONTEXT` can be inlined in both locations, unless it's providing you value in understanding the implementation as a hoisted value.

src/main-thread/main-thread.d.ts
```diff
@@ -14,10 +14,49 @@
...
+  getStorage(scope: string): { [key: string]: string };
+
+  /**
+   * Requests a change in localStorage or sessionStorage.
+   * @param scope 'local' or 'session'.
+   * @param key A storage item key. To change all keys, pass `null` here.
+   * @param value A storage value. To remove a key, pass `null` here.
+   * @return True if storage change was applied.
+   */
+  changeStorage(scope: string, key: string | null, value: string | null): boolean;
```
kristoferbaxter: Thanks for adding JSDoc comments to these methods!

src/main-thread/mutator.ts
```diff
@@ -63,33 +64,22 @@ export class MutatorProcessor {
     this.sanitizer = config.sanitizer;
     this.mutationPumpFunction = config.mutationPump;
 
-    const LongTaskExecutorInstance = LongTaskExecutor(stringContext, nodeContext, workerContext, objectContext, config);
-
+    const args = [stringContext, nodeContext, workerContext, objectContext, config];
```
kristoferbaxter: Makes sense given the current structure. I'm going to make some changes following this PR though to remove the duplication.

Also, please add a type to this definition.

src/main-thread/strings.ts
```diff
@@ -27,7 +27,7 @@ export class StringContext {
   /**
    * Return a string for the specified index.
    * @param index string index to retrieve.
-   * @returns string in map for the index.
+   * @return string in map for the index.
```
kristoferbaxter: Note for us in the future, let's have an ESLint rule or similar to enforce better JSDocs.

src/main-thread/worker.ts
```diff
@@ -38,11 +38,18 @@ export class WorkerContext {
...
     const { skeleton, strings } = createHydrateableRootNode(baseElement, config);
+
+    const cssKeys: Array<string> = [];
     for (const key in baseElement.style) {
-      keys.push(key);
+      cssKeys.push(key);
     }
+
+    // TODO(choumx): Sync read of all localStorage and sessionStorage a possible performance bottleneck?
+    const localStorageData = config.sanitizer ? config.sanitizer.getStorage('local') : window.localStorage;
```
kristoferbaxter: Likely a performance bottleneck, but let's keep moving forward.

We can followup with a benchmark using different sized versions of localStorage on a low-end device – based on the data we could pursue an async equivalent if needed.

src/test/main-thread/helpers/env.ts
```diff
@@ -24,7 +24,8 @@ export class Env {
   rafTasks: Array<Function>;
 
   constructor() {
-    this.jsdom = new JSDOM('<!DOCTYPE html>');
+    // JSDOM document URL necessary for localStorage access.
+    this.jsdom = new JSDOM('<!DOCTYPE html>', { url: 'http://localhost:3001' });
```
kristoferbaxter: Does this URL value need to be valid and active? Does it require SSL?

src/test/main-thread/test-main-thread.d.ts
```diff
@@ -17,6 +17,6 @@
 declare module 'jsdom' {
   export class JSDOM {
```
kristoferbaxter: I'll bet there is a great set of JSDOM definitions available for us to use instead of a local copy.

src/worker-thread/MutationTransfer.ts
```diff
@@ -26,6 +26,7 @@ import { Document } from './dom/Document';
 let pending = false;
 let pendingMutations: Array<number> = [];
 
+// TODO(choumx): Change `mutation` to Array<Uint16> to prevent casting errors e.g. integer underflow, precision loss.
```
kristoferbaxter: Good call. Looks like we are imprecise with the value types everywhere with Transferrables.

src/main-thread/commands/storage.ts
choumx: I'll inline it. 👍

src/main-thread/mutator.ts
choumx: Done.

src/worker-thread/Storage.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * Constructor is not directly invokable in the native implementation.
+   * @param document
+   * @param scope
+   */
+  constructor(document: Document, scope: string, data: { [key: string]: string }) {
+    this.document = document;
+    this.scope = scope;
+    this.data = data;
```
jridgewell: This **must must must** be a prototype-less object, or we're going to open ourselves up to all kinds of prototype injection bugs.

src/worker-thread/Storage.ts
```diff
@@ -0,0 +1,86 @@
...
+  }
+
+  get length() {
+    return Object.keys(this.data).length;
+  }
+
+  public key(n: number): string | null {
+    const keys = Object.keys(this.data);
+    const k = keys[n];
+    return k === undefined ? null : k;
```
jridgewell:
```suggestion
    return n >= 0 && n < keys.length ? keys[n] : null;
```

src/worker-thread/Storage.ts
```diff
@@ -0,0 +1,86 @@
...
+    transfer(this.document, [
+      TransferrableMutationType.STORAGE,
+      store(this.scope),
+      store(key),
+      0, // value == 0 represents deletion.
+    ]);
+  }
+
+  public clear(): void {
+    this.data = {};
```
jridgewell:
```suggestion
    this.data = Object.create(null);
```

src/test/mutation-transfer/storage.test.ts
```diff
@@ -0,0 +1,85 @@
...
+test.serial.cb('Storage.setItem', t => {
+  const { document, storage } = t.context;
+
+  expectMutations(document, (mutations, strings) => {
+    t.deepEqual(mutations, [TransferrableMutationType.STORAGE, strings.indexOf('scope'), strings.indexOf('foo'), strings.indexOf('bar')]);
+    t.end();
+  });
+
+  storage.setItem('foo', 'bar');
+});
```
jridgewell: We need a test that you can't poison `Object.prototype`.

src/main-thread/mutator.ts
```diff
@@ -63,33 +64,28 @@ export class MutatorProcessor {
     this.sanitizer = config.sanitizer;
     this.mutationPumpFunction = config.mutationPump;
 
-    const LongTaskExecutorInstance = LongTaskExecutor(stringContext, nodeContext, workerContext, objectContext, config);
-
+    const args: [StringContext, NodeContext, WorkerContext, ObjectContext, WorkerDOMConfiguration] = [
```
jridgewell: An options object would be cleaner for the processors to receive.

src/main-thread/commands/storage.ts
```diff
@@ -0,0 +1,84 @@
...
+          config.sanitizer.changeStorage(scope, key, value);
+        } else {
+          let storage;
+          if (scope == 'local') {
+            storage = window.localStorage;
+          } else if (scope == 'session') {
+            storage = window.sessionStorage;
+          }
+          if (storage) {
+            if (key != null) {
```
jridgewell: These `!=` with else clauses make my head hurt.

src/worker-thread/Storage.ts
choumx: Prototype poisoning is not a concern. Also author JS can already modify prototypes without this.

src/test/mutation-transfer/storage.test.ts
choumx: Ditto.

src/main-thread/commands/storage.ts
choumx: Done.

choumx:
Most comments should be addressed, PTAL.

src/worker-thread/Storage.ts
jridgewell: Poisoning might not be a concern, but this still produces incorrect results:

```js
localStorage.get('constructor'); // => function

PR #578, opened 2019-07-15
Unwrap author script

choumx:
Fixes issue where 3P libs set a property on the "window" and expect it to be available as a reference-less global. E.g.

```js
window.createjs = (function(){...})();

// Throws runtime error. 
createjs.foo();
```

This was also a suggested simplification by security since we're already deleting unwhitelisted globals in the "amp" variant.


src/worker-thread/index.amp.ts
```diff
@@ -210,31 +206,42 @@ function updateLongTask(global: WorkerGlobalScope) {
...
-      delete object[property];
-    } catch (e) {
-      console.warn(e);
+  /**
+   * @param object
+   * @param property
+   * @return True if property was deleted from object. Otherwise, false.
+   */
+  const deleteUnsafe = (object: any, property: string): boolean => {
+    if (WHITELISTED_GLOBALS.indexOf(property) < 0) {
```
kristoferbaxter: Let's take the opportunity to rename this while we're here, 'ALLOWLISTED_GLOBALS'

src/worker-thread/index.amp.ts
```diff
@@ -210,31 +206,42 @@ function updateLongTask(global: WorkerGlobalScope) {
...
+        return true;
+      } catch (e) {
+        console.warn(e);
+      }
     }
-  }
-
+    return false;
+  };
+  // Walk up global's prototype chain and dereference non-whitelisted properties
```
kristoferbaxter:
```suggestion
  // Walk up global's prototype chain and dereference non-allowlisted properties
```

src/worker-thread/index.amp.ts
```diff
@@ -210,31 +206,42 @@ function updateLongTask(global: WorkerGlobalScope) {
...
+    Object.getOwnPropertyNames(current).forEach(prop => {
+      if (deleteUnsafe(current, prop)) {
+        deleted.push(prop);
+      }
+    });
+    console.info(`Removed ${deleted.length} references from`, current, ':', deleted);
     current = Object.getPrototypeOf(current);
   }
-
+  // Wrap global.fetch() with our longTask API.
```
kristoferbaxter: Glad to see this comment, perhaps we should make it a bit more generic for future editors?

```suggestion
  // Wrap global methods as necessary
  // 1. global.fetch() with our longTask API.
```

kristoferbaxter:
A note for the future. This change starts the pattern of us precluding running multiple documents in a single Worker.

src/main-thread/worker.ts
```diff
@@ -37,44 +37,29 @@ export class WorkerContext {
...
+        self['localStorage'] = {}; // TODO(choumx): Remove.
+        var workerDOM = WorkerThread.workerDOM;
+        WorkerThread.hydrate(
+          workerDOM.document,
+          ${JSON.stringify(strings)},
+          ${JSON.stringify(skeleton)},
+          ${JSON.stringify(keys)},
+          [${window.innerWidth}, ${window.innerHeight}]);
+        workerDOM.document[${TransferrableKeys.observe}](this);
+        Object.keys(workerDOM).forEach(key => self[key] = workerDOM[key]);
```
kristoferbaxter: Shouldn't be an issue to do this, but we should verify that IE11 correctly handles this.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/keys#Notes

choumx: Here `workerDOM` is a WorkerDOMGlobalScope which I believe is a plain object. Definitely need to do an IE11 compatibility pass though e.g. `fetch` is not polyfilled.

src/worker-thread/canvas/CanvasRenderingContext2D.ts
```diff
@@ -85,29 +85,31 @@ export class CanvasRenderingContext2DShim<ElementType extends HTMLElement> imple
...
+    const isTestMode = !document.addGlobalEventListener;
 
     const upgradePromise = new Promise(resolve => {
       const messageHandler = ({ data }: { data: OffscreenCanvasToWorker }) => {
         if (
           data[TransferrableKeys.type] === MessageType.OFFSCREEN_CANVAS_INSTANCE &&
           data[TransferrableKeys.target][0] === canvas[TransferrableKeys.index]
         ) {
-          removeEventListener('message', messageHandler);
+          document.removeRemoveEventListener('message', messageHandler);
```
jridgewell: "removeRemove"? Why wasn't this caught by TypeScript?

choumx: Lol. Cause of this: https://github.com/ampproject/worker-dom/blob/9a1e6873ff84240ab0688b61c39df865233b1563/src/worker-thread/dom/Node.ts#L49-L50

Removing that breaks a lot of types but we probably should and typecast the callers as needed instead e.g. `(node as any)[property] = value`.

PR #627, opened 2019-08-22
Listen to global key events

### LINKED ISSUE #626: Support listening to global key events
When adding a `keypress` listener to the base element, the listener should be added to the document instead of the amp-script element (similar to the resize event handler). 

```
addEventListener('keypress', e => console.log(e.keyCode));
```

We might want to consider using this approach for all event listeners.

sebastianbenz:
Fixes #626


kristoferbaxter:
Closing in favor of #629.

PR #629, opened 2019-08-22
Global subscriptions

### LINKED ISSUE #626: Support listening to global key events

kristoferbaxter:
Global subscriptions (`addEventListener`) without a target, are safe to apply to the global target of the main thread since they cannot be prevent defaulted or have their propagation limited.

Fixes #626.


config/rollup.worker-thread.js
```diff
@@ -77,11 +77,11 @@ const ESModules = [
         ],
       }),
       replace({
-        WORKER_DOM_DEBUG: false,
+        WORKER_DOM_DEBUG: true,
```
kristoferbaxter: These changes made it possible to debug the issue. Looks like a bad merge made it so debugging was disabled on the debug output.

Issue #19, opened 2018-08-29
APIs needed by lit-html


justinfagnani:
We would love to get lit-html working in worker-dom.

At a glance it appears we need these APIs that aren't implemented yet:

- [ ] CustomElementRegistry#upgrade
- [ ] Document#adoptNode
- [x] Document#createComment
- [ ] Document#createTreeWalker
- [ ] Document#customElements
- [ ] Document#importNode
- [x] Element#innerHTML
- [ ] HTMLTemplateElement#content

`innerHTML` is going to be difficult, since we use the parsed tree immediately after setting it.

PR #631, opened 2019-08-25
Node constants

jdanyow:
Starting to add APIs needed for #19. Some of these constants are deprecated. Would you rather leave those out?

image
https://developer.mozilla.org/en-US/docs/Web/API/Node

### DIFF
--- a/package.json
+++ b/package.json
@@ -92,12 +92,12 @@
     {
       "path": "./dist/worker/worker.mjs",
       "compression": "brotli",
-      "maxSize": "11.2 kB"
+      "maxSize": "11.4 kB"
     },
     {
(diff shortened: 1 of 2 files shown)


kristoferbaxter:
Am curious, these values are already present in WorkerDOM but as a `const enum` ([TransferrableNodes.ts](https://github.com/ampproject/worker-dom/blob/master/src/transfer/TransferrableNodes.ts#L22)).

Is there a reason we need to expose them via `Node`?

jdanyow:
My bad, I didn't see the enum. The spec exposes these constants via the Node interface:
https://dom.spec.whatwg.org/#interface-node

PR #632, opened 2019-08-25
add NodeFilter

jdanyow:
Starting to add APIs needed for #19. Some of these constants are deprecated- should they be removed?
https://developer.mozilla.org/en-US/docs/Web/API/Document/createTreeWalker


kristoferbaxter:
I'd vote to keep them removed for now, but if/when someone asks for them add them back.

Issue #653, opened 2019-09-17
amp-list changes to input fields (i.e. autocomplete) aren't resolvable in amp-script


philkrie:
When amp-list is used to modify input fields (i.e. autocomplete), amp-script is unable to retrieve the modified input, and instead retrieves what the user typed.

You can see an example of the autocomplete in question [here](https://tryamp.msiteproject.com/bookit/index_test.html). If you type "PDX" into the To or From field, it autocompletes to "Portland, OR, OR, US". 

In the dev tools console, `document.getElementByID("to-input").value` resolves to `"Portland, OR, OR, US"`. In amp-script, it resolves to `"PDX"`.

(As noted in #628, amp-script has a limitation where you need to bind an event listener to the input field in order for it to access the value. This problem occurs with event listeners assigned to the autocompleted fields)

philkrie:
I've noticed this behaviour recently with other AMP components as well, such as amp-date-picker. If I assign amp-date-picker to output to a input[type="date"] field that is not hidden, amp-script is unable to retrieve the value. If the user clicks on the input field and inputs a date, then amp-script will get the value.

kristoferbaxter:
Leaving some raw notes here:

We might be able to resolve this by moving from an `oninput` property attached to anything that supports `change` events to using a mutation observer for the `value` attribute... which gets changed by `AMP.bind`.

```javascript
const observer = new MutationObserver(e => console.log(e.map(i => i.target.getAttribute('value'))));
observer.observe(el, {attributes: true});
```

We'll need to track the objects that have a `MutationObserver` attached to them, and remove the `MutationObserver`s when the `HTMLElement` is removed from the DOM.

```javascript
const observer = new MutationObserver((mutations, observer) => {
  if(mutations.type === 'childList' && //removed from DOM) {
    observer.disconnect();
  }
  console.log(mutations.map(i => i.target.getAttribute('value')));
});
observer.observe(el, {attributes: true, childList: true});

kristoferbaxter:
Early version here: https://github.com/ampproject/worker-dom/pull/665

kristoferbaxter:
Resolved with `0.21.0`.

PR #656, opened 2019-09-17
Default input listener value propagation

### LINKED ISSUE #628: Input's value doesn't resolve to anything unless I listed to "change"
```html
<amp-script layout="fixed-height" height="200" src="script.js">
  <div>
    <input type="text" id="input">
  </div>
  <p id="output"></p>
</amp-script>
```

If I do this:

```js
const output = document.querySelector('#output');
const input = document.querySelector('#input');

let c = 0;

setInterval(() => {
  output.textContent = `${input.value} ${c++}`;
}, 1000);
```

...then `input.value` is always empty.

However, adding `input.addEventListener('change', () => {})` to the code makes everything work as expected. It seems reading the value when no `change` event is registered means the value is never updated.

kristoferbaxter:
Resolves #628.

Tests are forthcoming.

### DIFF
--- a/demo/default-input-listener/author.js
+++ b/demo/default-input-listener/author.js
@@ -1,8 +1,13 @@
+const defaultInput = document.getElementById('default-syntax-input');
+const defaultOut = document.getElementById('default-syntax-output');
+
+setInterval(_ => (defaultOut.textContent = defaultInput.value), 500);
+
 const desc = document.createElement('p');
 const input = document.createElement('input');
 const valueOut = document.createElement('p');
 
-desc.textContent = 'Worker Thread Implementation';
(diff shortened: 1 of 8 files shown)


kristoferbaxter:
Hydrated elements are handled here: https://github.com/ampproject/worker-dom/pull/656/files#diff-d94e23bcb00127f9b39858f394b57831R58

Demo tests the usage.
1. In default HTML: https://github.com/ampproject/worker-dom/pull/656/files#diff-d1831d90d2e56aef28f2629cbc7ccc05R26
2. `.value` is monitored: https://github.com/ampproject/worker-dom/pull/656/files#diff-a219a008c8fa58957aa609f12627106cR4

philkrie:
Does this also resolve the issue described in issue #653?

Issue #369, opened 2019-03-16
Implement common GlobalEventHandlers e.g. onclick


choumx:
Related: #319 

Preact uses the presence of `'on' + event` for [type casing of registered event listeners](https://github.com/developit/preact/blob/3b448bb516922426fd48c10601c6b21990be380f/src/diff/props.js#L69-L81). 

E.g. if `onclick` is not implemented, then:

```jsx
<button onClick={handler()}>
```

...will result in:

```js
button.addEventListener('Click', ...); // Should be 'click'.
```

Event types are case-sensitive so this won't work.

React doesn't have this particular issue since it uses an [explicit mapping](https://github.com/facebook/react/blob/b87aabdfe1b7461e7331abb3601d9e6bb27544bc/packages/react-dom/src/events/SimpleEventPlugin.js#L143-L160) but it does do something similar for [isEventSupported](https://github.com/facebook/react/blob/b87aabdfe1b7461e7331abb3601d9e6bb27544bc/packages/react-dom/src/events/isEventSupported.js#L23-L38).

kristoferbaxter:
Resolved with #661.

PR #661, opened 2019-09-23
Global event handlers

kristoferbaxter:
Implement Global Event Handlers, frequently used by Frameworks to determine which events need special lower-casing of event names.

https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers

Implemented directly on `HTMLElement` since this appears to be the detection logic for Preact.


choumx:
Nice, this fixes #369.

PR #665, opened 2019-09-26
Mutation Observation

kristoferbaxter:
Use a MutationObserver to track attribute changes to `value` like `AMP.bind` uses when it changes a `value` property.


src/main-thread/commands/event-subscription.ts
```diff
@@ -39,6 +46,20 @@ export const applyDefaultInputListener = (workerContext: WorkerContext, node: Re
   shouldTrackChanges(node as HTMLElement) && node.oninput === null && (node.oninput = () => fireValueChange(workerContext, node));
 };
 
+/**
+ * Use a MutationObserver to capture value changes based on Attribute modification (frequently used by frameworks).
+ * @param worker whom to dispatch value toward.
+ * @param node node to listen to value changes on.
+ */
+export const applyDefaultMutationObservation = (workerContext: WorkerContext, node: RenderableElement): void => {
+  if (shouldTrackChanges(node as HTMLElement) && !!!monitoredNodes.get(node)) {
```
choumx: Should this be `!monitoredNodes.get(node)`?

src/main-thread/commands/event-subscription.ts
```diff
@@ -39,6 +46,20 @@ export const applyDefaultInputListener = (workerContext: WorkerContext, node: Re
   shouldTrackChanges(node as HTMLElement) && node.oninput === null && (node.oninput = () => fireValueChange(workerContext, node));
 };
 
+/**
+ * Use a MutationObserver to capture value changes based on Attribute modification (frequently used by frameworks).
+ * @param worker whom to dispatch value toward.
+ * @param node node to listen to value changes on.
+ */
+export const applyDefaultMutationObservation = (workerContext: WorkerContext, node: RenderableElement): void => {
```
choumx: Nit: Can we use a more descriptive name since this behavior is a bit non-intuitive? E.g. `sendValueChangeOnAttributeMutation`.

kristoferbaxter: Good idea, current name doesn't really explain it.

Issue #667, opened 2019-09-27
Regression in `HTMLInputElement.value = ''`


kristoferbaxter:
https://github.com/ampproject/amphtml/issues/24778

Mutation is correctly sent across from background thread to foreground, but is misapplied by the `PropertyProcessor`, which believes the new value is `null` not `''`.

kristoferbaxter:
Released with `v0.20.0`.

PR #669, opened 2019-09-27
Fix for Property Setters of empty strings not applying

kristoferbaxter:
Fix for #667.

Adds tests to prevent future regressions.


src/main-thread/commands/property.ts
```diff
@@ -21,17 +21,26 @@ import { NumericBoolean } from '../../utils';
...
         const targetIndex = mutations[startPosition + PropertyMutationIndex.Target];
         const target = nodeContext.getNode(targetIndex);
         const name = strings.get(mutations[startPosition + PropertyMutationIndex.Name]);
-        const isBooleanProperty = mutations[startPosition + PropertyMutationIndex.IsBoolean] === NumericBoolean.TRUE;
-        const value = isBooleanProperty
-          ? mutations[startPosition + PropertyMutationIndex.Value] === NumericBoolean.TRUE
-          : (mutations[startPosition + PropertyMutationIndex.Value] !== 0 && strings.get(mutations[startPosition + PropertyMutationIndex.Value])) ||
-            null;
+        const value = getValue(
+          mutations[startPosition + PropertyMutationIndex.IsBoolean] === NumericBoolean.TRUE,
```
wassgha: Just a suggestion but could this check for it being boolean be embedded into `getValue` so that the signature is `getValue(mutations, startPosition)`?

kristoferbaxter: Good call, changed over.

PR #697, opened 2019-10-18
New Feature `workerDOMPreventDefault` option

kristoferbaxter:
New option for `Node.addEventListener`'s `AddEventListenerOptions` object: `workerDOMPreventDefault`.

This is expected to be used when an event listener intends to `event.preventDefault()` unconditionally. The main thread uses the signal to perform this action unconditionally without a timing issue from the backend thread.

More tests are needed and examples before this is ready to submit.


src/main-thread/commands/event-subscription.ts
```diff
@@ -208,29 +235,29 @@ export const EventSubscriptionProcessor: CommandExecutorInterface = (strings, no
       return endPosition;
     },
     print(mutations: Uint16Array, startPosition: number, target?: RenderableElement | null): Object {
-      const addEventListenerCount = mutations[startPosition + EventSubscriptionMutationIndex.AddEventListenerCount];
-      const removeEventListenerCount = mutations[startPosition + EventSubscriptionMutationIndex.RemoveEventListenerCount];
-      const addEventListenersPosition = startPosition + EventSubscriptionMutationIndex.Events + removeEventListenerCount * EVENT_SUBSCRIPTION_LENGTH;
-      const endPosition =
-        startPosition + EventSubscriptionMutationIndex.Events + (addEventListenerCount + removeEventListenerCount) * EVENT_SUBSCRIPTION_LENGTH;
+      return {};
```
kristoferbaxter: TODO (@kristoferbaxter) Restore the Print method.

src/worker-thread/Event.ts
```diff
@@ -35,6 +35,13 @@ interface EventOptions {
 
 export type EventHandler = (event: Event) => any;
 
+export interface AddEventListenerOptions {
+  capture?: boolean;
+  once?: boolean;
+  passive?: boolean;
+  workerDOMPreventDefault?: boolean;
```
choumx: I understand the namespacing to prevent future collisions but "workerDOM" might be unfamiliar for users of `amp-script`. What do you think about using a different prefix e.g. `__` or `nonstandard`?

src/main-thread/commands/event-subscription.ts
```diff
@@ -188,17 +208,24 @@ export const EventSubscriptionProcessor: CommandExecutorInterface = (strings, no
...
       const removeEventListenerCount = mutations[startPosition + EventSubscriptionMutationIndex.RemoveEventListenerCount];
-      const addEventListenersPosition = startPosition + EventSubscriptionMutationIndex.Events + removeEventListenerCount * EVENT_SUBSCRIPTION_LENGTH;
+      const addEventListenersPosition =
+        startPosition + EventSubscriptionMutationIndex.Events + removeEventListenerCount * REMOVE_EVENT_SUBSCRIPTION_LENGTH;
       const endPosition =
-        startPosition + EventSubscriptionMutationIndex.Events + (addEventListenerCount + removeEventListenerCount) * EVENT_SUBSCRIPTION_LENGTH;
+        startPosition +
+        EventSubscriptionMutationIndex.Events +
+        addEventListenerCount * ADD_EVENT_SUBSCRIPTION_LENGTH +
+        removeEventListenerCount * REMOVE_EVENT_SUBSCRIPTION_LENGTH;
```
choumx: At this point, I wonder if it's simpler to have a separate mutator and message format for add vs. remove listener.

src/worker-thread/Event.ts
kristoferbaxter: I based this name off the prefix used by different browser vendors for nonstandard additions.

https://developer.mozilla.org/en-US/docs/Glossary/Vendor_Prefix

In this case it made sense to use `workerDOM` since it's the engine powering the various usecases like `amp-script`.

If this is confusing though, am happy to look at some alternatives.

src/main-thread/commands/event-subscription.ts
kristoferbaxter: Agree, it's getting a bit verbose. Will consider for a future PR.

PR #725, opened 2019-11-27
Remove Blob from Allowlisted Globals

kristoferbaxter:
Remove `Blob` from allowlist in the AMP output.


kristoferbaxter:
Good point, will take a look in a separate PR.

jridgewell:
`Response.prototype.blob` is still accessible via `fetch`

PR #764, opened 2020-01-17
Allow filtering mutation executions on user-visibility

choumx:
Enables enforcement user gesture constraints in use cases like `amp-script`.

Context: https://github.com/ampproject/amphtml/issues/26375

### DIFF
--- a/src/main-thread/mutator.ts
+++ b/src/main-thread/mutator.ts
@@ -113,8 +113,10 @@ export class MutatorProcessor {
    * Investigations in using asyncFlush to resolve are worth considering.
    *
    * TODO(amphtml): Consider returning the disallowed mutations for better error messaging.
+   *
+   * @param allowVisibleMutations
    */
-  private syncFlush = (allowVisibleMutations: boolean): void => {
+  private syncFlush = (allowVisibleMutations: boolean = true): void => {
     if (WORKER_DOM_DEBUG) {
       console.group('Mutations');
     }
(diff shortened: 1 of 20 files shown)


src/main-thread/mutator.ts
```diff
@@ -122,11 +126,13 @@ export class MutatorProcessor {
...
         const mutationType = mutationArray[operationStart];
+        // TODO(amphtml): Hoist `skip` up to entry point (index.amp.ts) for configurability?
+        const allow = allowVisibleMutations || !isUserVisibleMutation(mutationType);
         const executor = this.executors[mutationType];
         if (WORKER_DOM_DEBUG) {
-          console.log(ReadableMutationType[mutationType], executor.print(mutationArray, operationStart));
+          console.log(allow ? '' : '[disallowed]', ReadableMutationType[mutationType], executor.print(mutationArray, operationStart));
         }
-        operationStart = executor.execute(mutationArray, operationStart);
+        operationStart = executor.execute(mutationArray, operationStart, allow);
```
kristoferbaxter: Instead of preventing the mutation within each executor, could we stop execution here?

choumx: Yea, it's not the prettiest... we do need the return value of `execute()` for the next start position however.

I also tried adding a new interface method `nextPosition(): number` that just returns the next start position: 

```ts
if (allow) { 
  operationStart = executor.execute(mutationArray, operationStart);
} else {
  operationStart = executor.nextPosition(mutationArray, operationStart);
}
```

But that resulted in a lot of code duplication for more complex command executors, like here:
https://github.com/ampproject/worker-dom/blob/b83751455890349924a4c803fe4303ef33c802bd/src/main-thread/commands/event-subscription.ts#L209-L217

kristoferbaxter: Good point, didnt remember that detail.

PR #832, opened 2020-04-14
Do not strictly check force values in DOMTokenList.toggle

### LINKED ISSUE #831: DOMTokenList.toggle force attribute should not use strict equality
We are [doing strict equality checking](https://github.com/ampproject/worker-dom/blob/6c798ee908f08cc72502c0d0f441bb7413928743/src/worker-thread/dom/DOMTokenList.ts#L184) for the force attr, but it should be loose. 

```js
document.documentElement.classList.toggle('foo', '')
```
is adding a class in WorkerDOM, but it is not in the actual DOM.

patrickkettner:
fixes #831

[old spec discussion](https://github.com/whatwg/dom/issues/64)
relevant bits being gecko had a discrepancy (that is fixed in modern firefox), and IE ignores `force` all together, so this is never be good there either way

### DIFF
--- a/src/worker-thread/dom/DOMTokenList.ts
+++ b/src/worker-thread/dom/DOMTokenList.ts
@@ -170,19 +170,19 @@ export class DOMTokenList {
    * @param force changes toggle into a one way-only operation. true => token added. false => token removed.
    * @return true if the token is in the list following mutation, false if not.
    */
-  public toggle(token: string, force?: boolean): boolean {
+  public toggle(token: string, force?: any): boolean {
     if (WHITESPACE_REGEX.test(token)) {
       throw new TypeError('Uncaught DOMException');
     }
 
     if (!this[TransferrableKeys.tokens].includes(token)) {
-      if (force !== false) {
-        // Note, this will add the token if force is undefined (not passed into the method), or true.
+      if (force === undefined || !!force) {
+        // Note, this will add the token if force is undefined (not passed into the method), or truthy.
         this.add(token);
       }
       return true;
-    } else if (force !== true) {
-      // Note, this will remove the token if force is undefined (not passed into the method), or false.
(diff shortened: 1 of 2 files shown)


src/worker-thread/dom/DOMTokenList.ts
```diff
@@ -170,19 +170,19 @@ export class DOMTokenList {
...
-  public toggle(token: string, force?: boolean): boolean {
+  public toggle(token: string, force?: any): boolean {
     if (WHITESPACE_REGEX.test(token)) {
       throw new TypeError('Uncaught DOMException');
     }
 
     if (!this[TransferrableKeys.tokens].includes(token)) {
-      if (force !== false) {
-        // Note, this will add the token if force is undefined (not passed into the method), or true.
+      if (force === undefined || (force !== undefined && !!force)) {
```
kristoferbaxter, line 179: Might be missing something, but can this be accomplished with the following?

```suggestion
      if (force === undefined || !!force) {
```

src/worker-thread/dom/DOMTokenList.ts
```diff
@@ -170,19 +170,19 @@ export class DOMTokenList {
...
+        // Note, this will add the token if force is undefined (not passed into the method), or truthy.
         this.add(token);
       }
       return true;
-    } else if (force !== true) {
-      // Note, this will remove the token if force is undefined (not passed into the method), or false.
+    } else if (force === undefined || !force) {
```
kristoferbaxter, line 184: And here like this?

```suggestion
    } else if (!force) {
```

patrickkettner: totally right on both, just morning brain

patrickkettner:
updated!

Issue #783, opened 2020-01-29
Allow document.getElementsByName()


morsssss:
Maybe this isn't used that much as getting items by selector, ID, or classname... but it's a logical way to grab form inputs. If it's easy to implement, then could we? It would also mean we had support for the basic DOM elements getting methods:

* document.getElementById()
* document.getElementsByClassName()
* document.getElementsByName()
* document.getElementsByTagName()

instead of omitting one.

(see also #782)

kristoferbaxter:
Resolved with #833.

PR #833, opened 2020-04-14
Implement Element.getElementsByName

kristoferbaxter:
Implements `Element.getElementsByName`, resolving #783.

### DIFF
--- a/src/worker-thread/dom/Element.ts
+++ b/src/worker-thread/dom/Element.ts
@@ -480,8 +480,9 @@ export class Element extends ParentNode {
    * @param name value of name attribute elements must have to be returned
    * @return Element array with matching name attributes
    */
-  public getElementsByName(name: string): Array<Element> {
-    return matchChildrenElements(this, (element) => element.getAttributeNS(HTML_NAMESPACE, 'name') === name);
+  public getElementsByName(name: any): Array<Element> {
+    const stringName = '' + name;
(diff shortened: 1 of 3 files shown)


src/worker-thread/dom/Element.ts
```diff
@@ -452,29 +452,38 @@ export class Element extends ParentNode {
...
     );
   }
 
+  /**
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/Document/getElementsByName
+   * @param name value of name attribute elements must have to be returned
+   * @return Element array with matching name attributes
+   */
+  public getElementsByName(name: string): Array<Element> {
+    return matchChildrenElements(this, (element) => element.getAttributeNS(HTML_NAMESPACE, 'name') === name);
```
patrickkettner, line 484: nit: I don't think this would work as expected if someone passes `null`. The browser returns an empty NodeList, but I think this would return anything without a namespace, right?

patrickkettner:
other than the `null` case, lgtm

src/worker-thread/dom/Element.ts
kristoferbaxter: Good call out, I made a few changes based on this.

1. Moved from `name: string` to `name: any`.
2. Cast the `name` to a string via `'' + name`.
3. Added tests to validate its working as expected.

PR #850, opened 2020-04-23
Support callFunction API

samouri:
**summary**
The ability to invoke arbitrary worker functions from the main thread and get the result would be useful to two AMP use-cases, [Protocol Adapters](https://github.com/ampproject/amphtml/issues/26474) and [Allow real JS in amp-bind](https://github.com/ampproject/amphtml/issues/25684).

This PR exposes two new interfaces. A function named `callFunction` invoking functions on the worker, as well as a method for the worker code to mark which functions should be exposed to the main thread. I implement this by changing the return value from `upgradeElement` from a standard [Worker](https://developer.mozilla.org/en-US/docs/Web/API/Worker), into a wrapper object. Notably, it still delegates the `onerror` and `.terminate()` function calls to the underlying worker.

```js
// index.html
<div src="call-function.js" id="upgrade-me"></div>
<script type="module">
  import {upgradeElement} from '/dist/amp/main.mjs';
  const worker = await (upgradeElement(document.getElementById('upgrade-me'), '/dist/amp/worker/worker.mjs').)

   const result = await worker.callFunction('getRemoteData');
   console.log(result);
</script>

// worker.js
function getRemoteData() {
  return Promise.resolve({ big: 'tuna' });
}
exportFunction("getRemoteData", getRemoteData)
```

**questions i'd like help with**
1.  I noticed we don't specify `filesize` requirements for amp versions. Is this intentional / should I add them?

**todo / next steps**
- [ ] (Future PR) create a "lite" build for those that don't need to interact with DOM.
- [x] support passing arguments
- [x] add tests

cc @choumx @jridgewell


src/main-thread/worker.ts
```diff
@@ -82,6 +83,25 @@ export class WorkerContext {
...
     }
+
+    // TODO: How should this functionality be exposed?
+    // This is probably not the right place for it.
+    // Two ideas:
+    //   1. (this) a specialized function attached to the worker. Probably
+    //      would need to create a new type that extends the Worker interface.
+    //   2. a specific format of message devs can send via postMessage.
+    //      more of an untyped interface.
+    (this.worker as any).invokeFunction = (identifier: string, ...fnArguments: Array<any>) => {
```
choumx, line 94: > How should the functionality be exposed? A function attached to the worker is probably more ergonomic for callers than postMessage, but if anyone has better ideas/preferences I'm all ears!

Unless we have a really good reason to diverge, we should follow the `postMessage` convention here.

src/worker-thread/function.ts
```diff
@@ -0,0 +1,68 @@
...
+  if (msg[TransferrableKeys.type] !== MessageType.FUNCTION) {
+    return;
+  }
+
+  const functionMessage = msg as FunctionInvocationToWorker;
+  const fnIdentifier = functionMessage[TransferrableKeys.functionIdentifier];
+  const fnArguments = JSON.parse(functionMessage[TransferrableKeys.functionArguments]);
+  const index = functionMessage[TransferrableKeys.index];
+
+  const fn = (self as any)[fnIdentifier];
```
choumx, line 35: Let's consider adding a whitelisting scheme so only designated functions can be invoked. This would prevent accidentally calling DOM functions etc.

src/main-thread/worker.ts
samouri: The question was more about whether we has consumers of the Worker (amp-script) call `.postMessage` directly or if they should call something like `.callFunction`, which means exposing a non-standard interface on what was previously a normal [Worker](https://developer.mozilla.org/en-US/docs/Web/API/Worker)

src/worker-thread/function.ts
samouri: Lets talk about this. What do you mean? Should the person specify it in their worker?

e.g.:
```js
// worker.js
const EXPORTED_FUNCTIONS = [
  'func1',
  'func2'
]
```

It could make more sense for us to manually remove all the built-in functions instead and only expose the user added ones?

samouri:
To summarize some of the discussion from late last week with @choumx  and @jridgewell, here are the next steps:

1. Change the interface of `worker-dom` such that the API returns a new custom object with just `callFunction` and `terminate` as opposed to the standard Worker interface.
2. Bump version
3. TBD what to do about workers declaring exported functions.

src/worker-thread/function.ts
samouri: Added a worker global fn for declaring exports:`exportFunction(fnIdentifier: string, fn:Function)`

src/main-thread/commands/function.ts
```diff
@@ -0,0 +1,70 @@
+/**
+ * Copyright 2019 The AMP HTML Authors. All Rights Reserved.
```
choumx, line 2: Nit: 2020

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ampproject/worker-dom",
-  "version": "0.24.0",
+  "version": "1.24.0",
```
choumx, line 3: Let's just bump minor version.

src/main-thread/commands/function.ts
```diff
@@ -0,0 +1,70 @@
...
+  let resolve;
+  let reject;
+  const promise = new Promise((res, rej) => {
+    resolve = res;
+    reject = rej;
+  });
+
+  const index = fnCallCount;
+  promiseMap[index] = { resolve, reject, promise };
+  return { promise, index: fnCallCount++ };
```
choumx, line 33: Do we need to worry about overflow?

src/main-thread/commands/function.ts
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import { CommandExecutorInterface } from './interface';
+import { TransferrableMutationType, FunctionMutationIndex } from '../../transfer/TransferrableMutation';
+import { ResolveOrReject } from '../../transfer/Messages';
+
+let fnCallCount = 0;
+const promiseMap: { [id: number]: any } = {};
```
choumx, line 22: Please add comments describing what these module-level states hold.

src/main-thread/exported-worker.ts
```diff
@@ -0,0 +1,55 @@
...
+ * limitations under the License.
+ */
+
+import { WorkerContext } from './worker';
+import { WorkerDOMConfiguration } from './configuration';
+import { registerPromise } from './commands/function';
+import { FunctionCallToWorker, MessageType } from '../transfer/Messages';
+import { TransferrableKeys } from '../transfer/TransferrableKeys';
+
+export class ExportedWorker {
```
choumx, line 23: Please add a class comment.

src/main-thread/exported-worker.ts
```diff
@@ -0,0 +1,55 @@
...
+  workerContext_: WorkerContext;
+  config_: WorkerDOMConfiguration;
```
choumx, line 25: I don't think we use trailing underscores in the rest of the code.

src/main-thread/configuration.ts
```diff
@@ -46,6 +46,8 @@ export interface InboundWorkerDOMConfiguration {
   hydrateFilter?: HydrationFilterPredicate;
   // Executor Filter, allow list
   executorsAllowed?: Array<number>;
+  // Enable callFunction API
+  callFunctionAllowed?: boolean;
```
choumx, line 50: Is this necessary since we already check `executorsAllowed`?

morsssss:
Coolness

/cc @AndrewKGuan @alankent

package.json
samouri: https://0ver.org/

src/main-thread/commands/function.ts
samouri: I don't think so. The `fnCallCount` is only used to correlate fn call requests and its associated response. Reusing an older index is fine as long as the reused index had actually completed.

It would technically be a bug if there were some super super long running task with `index=0` and then while it were running, we launched `Number.MAX_VALUE` other calls until overflow. And then the next call would overwrite the first one.


Seems okay to me?

samouri: TIL js doesn't overflow integers! It rounds to either `Number.MAX_VALUE` or to `Infinity`.  

For example, `Number.MAX_VALUE + 1 === Number.MAX_VALUE`

src/main-thread/exported-worker.ts
samouri: removing 👍

src/main-thread/configuration.ts
samouri: good point, removing

src/main-thread/commands/function.ts
choumx: Me neither but should be easy to wrap to 0 when we hit MAX_VALUE.

src/main-thread/commands/function.ts
```diff
@@ -0,0 +1,87 @@
...
+    execute(mutations: Uint16Array, startPosition: number): number {
+      if (allowedExecution) {
+        const status = mutations[startPosition + FunctionMutationIndex.Status];
+        const index = mutations[startPosition + FunctionMutationIndex.Index];
+        const value = mutations[startPosition + FunctionMutationIndex.Value];
+
+        if (status === ResolveOrReject.RESOLVE) {
+          promiseMap[index].resolve(JSON.parse(strings.get(value)));
+        } else {
+          promiseMap[index].reject(JSON.parse(strings.get(value)));
```
choumx, line 66: How is `undefined` return value handled here? Are there other cases where JSON.parse can fail/throw? 

Nit: Hoist the `JSON.parse` bit.

src/test/main-thread/commands/function.test.ts
```diff
@@ -0,0 +1,71 @@
+/**
+ * Copyright 2019 The AMP HTML Authors. All Rights Reserved.
```
choumx, line 2: Nit: 2020

src/worker-thread/function.ts
```diff
@@ -0,0 +1,83 @@
...
+        transfer(document, [
+          TransferrableMutationType.FUNCTION_CALL,
+          ResolveOrReject.REJECT,
+          index,
+          store(JSON.stringify(`[worker-dom]: Function "${fnIdentifier}" threw: "${errorMessage}"`)),
+        ]);
+      },
+    );
+}
+export function exportFunction(name: string, fn: Function) {
```
choumx, line 66: Optional: An "unregister" counterpart is typical.

src/main-thread/commands/function.ts
samouri: sure

src/main-thread/commands/function.ts
samouri: Answer: it breaks :). Solving it required a bit of a hack since `StringContext` will return `''` in the case of a falsy value. Therefore I added a new `hasIndex` method so we can distinguish between undefined and the empty string.

src/worker-thread/function.ts
samouri: What would the purpose be here? We can safely unregister without the caller's input, so I don't think we need to expose that.

src/main-thread/commands/function.ts
choumx: Did you push changes yet?

samouri: https://i.giphy.com/media/yxF4HmDIXw83S/giphy.webp

src/worker-thread/function.ts
choumx: Niche usage pattern probably but would allow the registered function to be garbage collected.

samouri: I'm still unsure what you mean. The only usage pattern I could think of would be if the client wants to "cancel" an in-progress promise. But in reality they can't actually cancel it so it is effectively the same as ignoring the returned promise which is less work anyway.

samouri:
@choumx: can you merge? I don't have access :)

kristoferbaxter:
Added you to the write permission group @samouri.

src/worker-thread/function.ts
```diff
@@ -0,0 +1,83 @@
...
+  const fnArguments = JSON.parse(functionMessage[TransferrableKeys.functionArguments]);
+  const index = functionMessage[TransferrableKeys.index];
+
+  const fn = exportedFunctions[fnIdentifier];
+  if (!fn) {
+    transfer(document, [
+      TransferrableMutationType.FUNCTION_CALL,
+      ResolveOrReject.REJECT,
+      index,
+      store(JSON.stringify(`[worker-dom]: Exported function "${fnIdentifier}" could not be found.`)),
```
kristoferbaxter, line 43: Is there a reason we need to do this? Could we just reject with a generic error?

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ampproject/worker-dom",
-  "version": "0.24.0",
+  "version": "0.25.0",
```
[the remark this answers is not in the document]
samouri: how can I automatically update the version?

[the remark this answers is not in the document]
kristoferbaxter: The release script will do this automatically (check out RELEASING.md).

src/worker-thread/function.ts
choumx: I was thinking like: 

```js
self.exportFunction('oneAndDone', () => {
  // Do some one-time work.
  self.unexportFunction('oneAndDone'); // Release reference to this function.
});
```

samouri: ahh, when you said unregister I was thinking about `unregisterPromise` instead of `unexportFunction`.  Being able to unexport functions sounds like a footgun to me, I'd rather hold off unless theres a decent use case for it.

src/worker-thread/function.ts
samouri: Isn't it more helpful to give the devs a useful error message?

demo/call-function/index.html
```diff
@@ -0,0 +1,51 @@
...
+
+        worker.callFunction('returnsUndefined').then(result => {
+          console.log(`undefined --> ${result}`)
+        })
+
+        worker.callFunction("reject").catch((err) => console.error(err));
+        worker
+          .callFunction("tooCoolToExist")
+          .catch((err) => console.error(err));
+        worker
```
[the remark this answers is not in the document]
samouri: sure thing!

src/worker-thread/function.ts
samouri: Discussed this in Chat. The current method has higher DevEx whereas not having a helpful error message would save bytes.

We can have the best of both worlds by eventually having two builds, one for dev and one for prod. For now I'd like to merge with the slightly higher DevEx option esp as we work on getting to the optimal solution.

samouri:
I'm going to hold off on merging this until I land a fix for https://github.com/ampproject/worker-dom/pull/872

PR #855, opened 2020-04-29
[RFC] Add npm library mode

mizchi:
This PR is proposal to add `npm library mode` build.

## API Proposal

It generates them.

- `worker-dom/dist/lib/main.mjs`:  export `attachWorker(element: HTMLElement, worker: Worker): void`
- `worker-dom/dist/lib/worker.mjs`: export `ready: Promise<void>` (DOM is ready to touch)

## Background

I think worker-dom can be general purpose library for frontend performance, but today's api does not fit webpack. Under using webpack, there are many ways to bundle worker code like `worker-plugin` and `worker-loader`.

This proposal provides seamless way with webpack.

## Example

with `worker-plugin` (https://github.com/GoogleChromeLabs/worker-plugin)

```js
// index.js
import { attachWorker } from '@ampproject/worker-dom/dist/lib/main.mjs';

const el = document.createElement('main');
el.innerHTML = `
<div>
  <h1>main</h1>
  <button>click</button>
</div>
`;
document.body.appendChild(el);

// bundle worker code with worker-plugin
const worker = new Worker('./worker.js', { type: 'module' });

attachWorker(el, worker);
```

```js
// worker.js
import { ready } from '@ampproject/worker-dom/dist/lib/worker.mjs';

async function run() {
  await ready;

  // dom mutation
  const el = document.createElement('h2');
  el.textContent = 'hello from worker';
  document.body.appendChild(el);

  // handler
  const button = document.querySelector('button');
  button.onclick = (ev) => {
    console.log('button:onclick', ev);
  };
}

run();
```

See full code `demo/as-webpack-lib`

## Working Demo

https://worker-dom-as-lib-example.netlify.app/

sample repository (use published `@mizchi/worker-dom`)

https://github.com/mizchi/worker-dom-webpack-example

https://wizardly-benz-02c23e.netlify.app/


src/main-thread/lib.ts
```diff
@@ -0,0 +1,138 @@
...
+    );
+
+    if (config.onReceiveMessage) {
+      config.onReceiveMessage(message);
+    }
+  };
+}
+
+// TODO: Refactor with original
+class WorkerContextFromInstance {
```
mizchi, line 64: WIP: I copied it from WorkerContext and patched. Is there clear way?

src/main-thread/lib.ts
```diff
@@ -0,0 +1,138 @@
...
+    worker.postMessage({
+      __init__: [
+        strings,
+        skeleton,
+        cssKeys,
+        globalEventHandlerKeys,
+        [window.innerWidth, window.innerHeight],
+        {}, // localStorage
+        {}, // sessionStorage
+      ],
```
mizchi, line 108: Maybe adding `TransferrableKeys` and some types like `Initializer` are better but at first makes it simple

powerivq:
We are closing this PR for now due to being inactive. Please feel free to reopen.

PR #873, opened 2020-06-08
Log less in AMP output

kristoferbaxter:
Resolves https://github.com/ampproject/amphtml/issues/28714

### DIFF
--- a/src/worker-thread/index.amp.ts
+++ b/src/worker-thread/index.amp.ts
@@ -61,6 +61,8 @@ import { initialize } from './initialize';
 import { wrap as longTaskWrap } from './long-task';
 import { callFunctionMessageHandler, exportFunction } from './function';
 
+declare const WORKER_DOM_DEBUG: boolean;
+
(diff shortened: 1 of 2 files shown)


src/worker-thread/index.amp.ts
```diff
@@ -61,6 +61,8 @@ import { initialize } from './initialize';
 import { wrap as longTaskWrap } from './long-task';
 import { callFunctionMessageHandler, exportFunction } from './function';
 
+declare const WORKER_DOM_DEBUG: boolean;
```
samouri, line 64: is there any location we could put this such that we'd only need to declare it once?

kristoferbaxter: Good point, I'll address.

kristoferbaxter: Moved to `worker-thread.d.ts` like defined for `main-thread.d.ts`.

Perhaps in the future we should move this to a common definition file for both threads.

samouri, line 64: I think you forgot to delete this line

PR #881, opened 2020-07-15
Remove notion of allowedExecutors in the config. 

samouri:
Feel free to let me know that this this is a desired functionality. I couldn't find it being used in any codepaths.

### DIFF
--- a/src/main-thread/commands/attribute.ts
+++ b/src/main-thread/commands/attribute.ts
@@ -14,7 +14,7 @@
  * limitations under the License.
  */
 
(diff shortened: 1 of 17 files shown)


samouri:
@choumx: this is currently not being piped through to amp-script in any way. We also don't have any plans to add in executors that amp-script should ban, and even if we do, we should likely exclude them from the binary instead of skipping their execution (save bytes).

samouri:
amp-script doesn't have any APIs it would like to block yet, meaning even more unused bytes

PR #882, opened 2020-07-15
Introduce "nodom" binary

samouri:
**summary**
Interactive pieces of AMP would like to safely run arbitrary JS without impacting the main thread. While often that may involve dom mutation, there are instances where that isn't necessary. A current example is [Protocol Adapters](https://github.com/ampproject/amphtml/issues/26474). 

For cases like these, it doesn't make sense to ship a binary containing the vdom implementation. Instead we could ship a dom-less "lite" version of `worker-dom`.

- For the main-thread code, this means removing all of the unnecessary Executors.
- For worker-thread code, this means ensuring `Element` and friends don't enter the build. 

The method for removing the unnecessary code was similar in both cases. 
Namely, hoisting the relevant to the entrypoints.

**results (mjs)**

|          | main-thread | worker-thread |
|----------|-------------------|---------------------|
| AMP      | 11.24 kb           | 31.04 kb               |
| AMP nodom | 7.14 kb            | 1.45 kb                |

Note: using the `rollup-plugin-analyzer` I was able to detect further improvements we could make to main-thread code. worker-thread seemed relatively optimized. I plan on adding this as a dev dependency in a future PR.


src/main-thread/configuration.ts
```diff
@@ -84,13 +100,42 @@ export interface WorkerDOMConfiguration {
...
     {
       mutationPump: requestAnimationFrame.bind(null),
       executorsAllowed: DefaultAllowedMutations,
+      getExecutors,
     },
     config,
   );
 }
+
+export function getLiteProcessors() {
```
samouri, line 117: I'm not certain if this gets DCEd the way I'd expect. May need to move these two declarations into separate files used by the entrypoints

samouri:
Based on the ease of integration with `amp-script`, I'm going to split out the `main-thread` optimization into its own PR. Mainly since I'm unsure how possible it is to take advantage of it.

src/worker-thread/index.amp.ts
```diff
@@ -60,96 +60,10 @@ import { Text } from './dom/Text';
 import { initialize } from './initialize';
 import { wrap as longTaskWrap } from './long-task';
 import { callFunctionMessageHandler, exportFunction } from './function';
+import { deleteGlobals } from './index.lite.amp';
```
choumx, line 63: Why put this in the new entry point rather than this entry point or a shared file?

samouri: Shared file would work. I was worried that if "lite" imported from the main one it could drag in more unnecessary bytes.

samouri: Actually, since the entrypoints are automatically going to have side effects the only correct thing to do is pull out into a share file.

samouri:
Screen Shot 2020-07-20 at 12 45 09 PM

kristoferbaxter:
Quick comment while out on leave: Please verify the contents of these new binaries, since the `mjs` output should always be smaller than the `js` output.

samouri:
> Quick comment while out on leave: Please verify the contents of these new binaries, since the mjs output should always be smaller than the js output.

Good to know! I didn't know if the lack of `compilePlugin` was intentional for the amp mjs binary or not. I'm assuming not then :). Fixing here: https://github.com/ampproject/worker-dom/pull/885

samouri:
> Quick comment while out on leave: Please verify the contents of these new binaries, since the `mjs` output should always be smaller than the `js` output.

@kristoferbaxter:  seems like the `mjs` binaries are actually supposed to be bigger for AMP. The mjs builds are used for dev and the js for prod builds (.max.js vs. .min.js).

src/worker-thread/dom/DocumentLite.ts
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+
+import { PostMessage } from '../worker-thread';
+import { Phase } from '../../transfer/Phase';
+import { TransferrableKeys } from '../../transfer/TransferrableKeys';
+import { set as setPhase } from '../phase';
+import { WorkerDOMLiteGlobalScope } from '../WorkerDOMGlobalScope';
+
+export class DocumentLite {
```
choumx, line 23: Let's name this `DocumentStub`.

src/worker-thread/WorkerDOMGlobalScope.ts
```diff
@@ -118,3 +119,9 @@ export interface WorkerDOMGlobalScope extends GlobalScope {
   addEventListener: (type: string, handler: EventHandler) => void;
   removeEventListener: (type: string, handler: EventHandler) => void;
 }
+
+export interface WorkerDOMLiteGlobalScope {
```
choumx, line 123: `WorkerNoDOMGlobalScope`

src/worker-thread/delete-globals.ts
```diff
@@ -0,0 +1,124 @@
+const ALLOWLISTED_GLOBALS: { [key: string]: boolean } = {
```
choumx, line 1: Missing copyright header.

choumx, line 1: This file is only used by AMP entry points -- let's put it in the `amp/` folder.

src/worker-thread/WorkerDOMGlobalScope.ts
samouri: 👍

PR #883, opened 2020-07-16
dev deps: add rollup-plugin-analyzer

samouri:
makes it slightly easier to toss this in temporarily to rollup configs.


package.json
```diff
@@ -64,6 +64,7 @@
     "prettier": "2.0.5",
     "rimraf": "3.0.2",
     "rollup": "2.15.0",
+    "rollup-plugin-analyzer": "^3.3.0",
```
choumx, line 67: This should be exact; no `^`.

PR #884, opened 2020-07-20
Polyfill Request and Cancel Animation Frame for Worker Thread

kristoferbaxter:
Some browsers do not ship this API to worker threads.

### DIFF
--- a/package.json
+++ b/package.json
@@ -90,10 +90,10 @@
       "./dist/**/*.mjs"
     ],
     "./dist/worker/worker.mjs": {
-      "brotli": "11.5 kB"
+      "brotli": "11.6 kB"
     },
     "./dist/worker/worker.js": {
-      "brotli": "13 kB"
+      "brotli": "13.5 kB"
     },
     "./dist/main.mjs": {
       "brotli": "4 kB"
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ampproject/worker-dom",
-  "version": "0.25.0",
+  "version": "0.25.1",
   "description": "A facsimile of a modern DOM implementation intended to run in a Web Worker.",
   "main": "dist/main",
   "files": [
--- a/src/worker-thread/AnimationFrame.ts
+++ b/src/worker-thread/AnimationFrame.ts
@@ -20,54 +20,61 @@ interface Queued {
   cancelled: boolean;
 }
 
-export const { rafPolyfill, cafPolyfill }: { rafPolyfill: typeof requestAnimationFrame; cafPolyfill: typeof cancelAnimationFrame } = (() => {
-  const frameDuration = 1000 / 60;
-  let last: number = 0;
-  let id: number = 0;
-  let queue: Array<Queued> = [];
+const frameDuration = 1000 / 60;
+let last: number = 0;
+let id: number = 0;
+let queue: Array<Queued> = [];
 
-  function scheduleNext() {
-    var _now = performance.now(),
-      next = Math.max(0, frameDuration - (_now - last));
-    last = next + _now;
-    setTimeout(function () {
-      var cp = queue.slice(0);
-      // Clear queue here to prevent
-      // callbacks from appending listeners
-      // to the current frame's queue
-      queue.length = 0;
-      for (var i = 0; i < cp.length; i++) {
-        if (!cp[i].cancelled) {
-          try {
-            cp[i].callback(last);
-          } catch (e) {
-            setTimeout(function () {
-              throw e;
-            }, 0);
-          }
-        }
+/**
+ * Schedules the accumulated callbacks to be fired 16ms after the last round.
+ */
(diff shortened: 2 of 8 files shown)


src/worker-thread/AnimationFrame.ts
```diff
@@ -0,0 +1,76 @@
...
+  caf = cancelAnimationFrame;
+}
+
+interface Queued {
+  handle: number;
+  callback: Function;
+  cancelled: boolean;
+}
+
+if (typeof requestAnimationFrame === 'undefined') {
```
choumx, line 31: Nit: Instead of conditionally exporting the polyfill, I'd consider always exporting the polyfill and conditionally setting it on the global scope object. Would make testing easier.

src/worker-thread/AnimationFrame.ts
```diff
@@ -0,0 +1,76 @@
...
+              setTimeout(function () {
+                throw e;
+              }, 0);
+            }
+          }
+        }
+      }, Math.round(next));
+    }
+    queue.push({
+      handle: ++id,
```
choumx, line 62: Nit: overflow not handled.

src/worker-thread/AnimationFrame.ts
```diff
@@ -0,0 +1,76 @@
...
+      callback: callback,
+      cancelled: false,
+    });
+    return id;
+  };
+
+  caf = function (handle) {
+    for (var i = 0; i < queue.length; i++) {
+      if (queue[i].handle === handle) {
+        queue[i].cancelled = true;
```
choumx, line 72: Nit: Should `break;` here, or use a helper for `Array.find`-like functionality if we have one (no IE11).

src/worker-thread/AnimationFrame.ts
kristoferbaxter: Agree. Good idea.

src/worker-thread/AnimationFrame.ts
kristoferbaxter: Agree here too, this is a generic polyfill and should be refined.

src/worker-thread/AnimationFrame.ts
```diff
@@ -0,0 +1,76 @@
...
+  const frameDuration = 1000 / 60;
+  let last: number = 0;
+  let id: number = 0;
+  let queue: Array<Queued> = [];
+
+  raf = function (callback) {
+    if (queue.length === 0) {
+      var _now = performance.now(),
+        next = Math.max(0, frameDuration - (_now - last));
+      last = next + _now;
```
samouri, line 41: Wouldn't it be more accurate if we recorded `last` immediately after the last cb fired?

samouri: Lets say a frame takes 30ms to render, do browsers schedule the next animation frame for 2ms later, or do they wait another 16ms

samouri: answer via will: schedules for after another 16ms.

samouri: actually, unless done very carefully that would cause a weird race if someone calls `rAF` within an `rAF` cb

samouri: Ah! I tested this out via an experiment, and it turns out that the next frame would happen after 2ms. I think the logic is that `rAF` attempts to sync with the monitor refresh rate.

If interested, run this in the console of a new tab:
```js
requestAnimationFrame(() => {
    let start = performance.now()
    requestAnimationFrame(() => {
      // Spend too much perf budget in a rAF
      while( performance.now() - start < 31) {}
      let t2 = performance.now()
      // Now that we spent 31ms in a frame, how long till the next one
      requestAnimationFrame(() => {
        let end = performance.now()
        console.log(`Total time: ${end - start}, time since the end of the last frame: ${end - t2}`)
      })
    })
});
```

samouri:
I've made all of the highlighted changes and pushed to this branch.
Have a couple of items I'm still working through:

1. How to catch the "uncaught exception" within tests since the rAF throws them out-of-band by design.
2. How thoroughly should we test the polyfill. I.e should I install `sinon/fake-timers` in order to precisely test callback timing.

samouri:
> A fake timer stubs setTimeout so you could do the same and expect a throw.

Done! 

> Would be nice to verify the callback happens within ~1 frame. Also helps with above.

I added two tests.
1. Callbacks happen in the same tick
2. raf within a raf waits for the next batch

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ampproject/worker-dom",
-  "version": "0.25.0",
+  "version": "0.25.1",
```
samouri, line 3: I was under the impression the release tool automatically handled version bumping. Should I revert this back to 0.25.0?

kristoferbaxter: It's fine.

PR #885, opened 2020-07-20
worker-thread amp binary: add in missing compiler plugin

samouri:
Should we be running the AMP version through terser? If so then I'll also add this to main-thread.

**before**: mjs was +17kb 
before

**after**: mjs is -1.62kb
after


samouri:
Closing as this was intentional: https://github.com/ampproject/worker-dom/pull/882#issuecomment-661322037

Issue #217, opened 2019-01-02
Question: Will this also support IE11?


himhim1125:
I have tried to run the demo in local but find that most demos are not working well when using IE11

kristoferbaxter:
Yes, this will support IE11. Most of the missing pieces at this point are listed as TODOs in the codebase.

PR #886, opened 2020-07-22
ie11 support for nomodule binaries

samouri:
**summary**
Addresses https://github.com/ampproject/worker-dom/issues/217

There were a few changes necessary to make nomodule binaries IE11 compat:
1. Polyfilling newer ECMAScript features via Babel's `useBuiltIns` + core-js.
2. Fix illegal spread within DomTokenList (attempt to spread a string)
3. Convert `node.remove()` to `node.parentNode.removeChild(node)` within ChildList
4. Replaced usage of fetch with XHR.

**testing**

I converted two of the examples from `demo`, "hello-world" and "preact todomvc"  to be ie11 compatible. An interesting note here is that we can't prettify the source, since prettier will format in breaking ways.

**open questions**
- For the [babelHelpers](https://github.com/rollup/plugins/tree/master/packages/babel#babelhelpers) I am using "bundled" which is meant for applications as opposed to "runtime" mode meant for libs. This means that you can use `worker-dom` nomodule binaries _without_ a bundler, although introduces the possibility of double-polyfilling for those that do use bundlers. Any advice for how to proceed here would be much appreciated!
- ~The fetch polyfill is heavy handed and lumped in with all nomodule builds as opposed to be conditionally fetched. One option to fix that is if we compiled a separate binary and add a script tag for the polyfill in case we need to. Is using a CDN fine for this option or should we provide the bits ourselves? Is this a reasonable approach?~ Opting for removing singular usage of `fetch` in favor of xhr.


config/rollup.plugins.js
```diff
@@ -28,19 +28,22 @@ const walk = require('acorn-walk');
  * - allowPostMessage Should we allow postMessage to/from the Worker?
  */
 export function babelPlugin({ transpileToES5, allowConsole = false, allowPostMessage = true }) {
-  const targets = transpileToES5 ? { browsers: ['last 2 versions', 'ie >= 11', 'safari >= 7'] } : { esmodules: true };
+  const targets = transpileToES5 ? { browsers: ['last 2 versions', 'ie >= 10', 'safari >= 7'] } : { esmodules: true };
```
samouri, line 31: note: `last 2 versions` already meant ie10 (I noticed this in debug logs). If we want to exclude ie10 we'll need to change from last 2 versions as well.

samouri:
@kristoferbaxter: I'm marking this as ready for review but its not yet ready for merge. I'd like help answering some of the open questions I wrote out above. 

The good news: it works!
The bad news: bundle size for .js goes through the roof

samouri:
other bad news: we need to manually polyfill the main-thread-amp code within amp-script.
after talking with Justin, looks like a really good next step is to attempt to pare down the necessary transforms. `{debug: true}` mode for preset-env will be invaluable and we can 1 by 1 modify try to remove usages of polyfilled bits and/or come up with lighterweight polyfills (non spec compliant)

config/rollup.plugins.js
kristoferbaxter: Good point, should we just specify `ie >= 11`?

samouri: I opted to be more verbose/specific. switching to just `ie 11` could work too

samouri:
Removing this, considering: https://github.com/ampproject/amphtml/issues/34453

PR #913, opened 2020-08-20
Support Dirty Flag for HTMLInputElement.value

kristoferbaxter:
Demonstrates how to fix this issue by relying on the `Attribute` before the setter was called on the `Property`.

Likely we should have a generic pattern for this.


samouri:
Note: this fixes https://github.com/ampproject/worker-dom/pull/913

samouri:
Seems to be an issue with the `np` tarball in package.json:

Screen Shot 2020-08-25 at 4 24 59 PM

kristoferbaxter:
We've been relying on a custom version of `np` due to breaking changes in the main line. Looks like time has caught up to us.

samouri:
Attempting to update here: https://github.com/ampproject/worker-dom/pull/915, if that goes through I'll rebase this branch.

samouri:
@kristoferbaxter: this is ready for a review when you have time. Not sure if you intended to merge the new demo files or were just using them for manual testing, my preference would be to delete before a merge.

kristoferbaxter:
Looks like we went over the size limit, taking a look.

PR #915, opened 2020-08-26
update to latest version of np

samouri:
@kristoferbaxter: do you remember what broke in the mainline of `np`, so I can check if its been fixed?

### DIFF
--- a/package.json
+++ b/package.json
@@ -55,7 +55,7 @@
     "cross-env": "7.0.2",
     "esm": "3.2.25",
     "husky": "4.2.5",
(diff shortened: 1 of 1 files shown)


samouri:
Updating `jsdom` as well, since I believe we ran in to: https://github.com/jsdom/whatwg-url/issues/166

Screen Shot 2020-08-26 at 1 18 41 PM

kristoferbaxter:
The issue was related to 2fa and publishing. With 2fa enabled, publishing wouldn't work.

Issue #917, opened 2020-08-27
Cant use HTMLInputElement.click()


ayobami303:
I am trying to use HTMLInputElement.click() within amp-script.
`const btn = document.querySelector('#mybutton');
  btn.click()`

I am getting this error:

[amp-script] Uncaught TypeError: btn.click is not a function

samouri:
Thank you for reporting this @ayobami303. I'll work on a fix.

xaander1:
@samouri 
The issue is not fully fixed. 
I just tested out using a **link** it doesn't work.

```
const input = document.getElementById("myinput")
const mylink = document.getElementById("myDynamicUrl")

input.addEventListener("keyup", function(event) {
    if (event.keyCode === 13) {
        mylink.click()
        console.log("Enter pressed")
    }
})
```

> Buttons work just fine.

samouri:
I'm surprised! Will try to take a look next week

samouri:
Confirmed that `.click()` is broken for anchor tags. Can't figure out a simple way to make the Worker navigate page, so will need to perform it via postMessage.

memerapi:
hi @samouri ,
can you give an example on how to navigate page via postMessage?

samouri:
@memerapi: by postMessage I was talking about an implementation detail of how .click() would work. I don't know if there is a way for you to fix it on your end without submitting a PR to worker-dom (or writing your own listener on the main thread for worker messages)

memerapi:
looks like i misinterpreted it, that's makes more sense now.
thanks @samouri

PR #919, opened 2020-08-27
HTMLElement.click() support

samouri:
Fixes https://github.com/ampproject/worker-dom/issues/917


samouri:
> Not sure if this in the compatibility table, but worth a look.

It is and already has a check mark 🙃

PR #926, opened 2020-09-03
add compilePlugins to dist/amp/worker/worker.nodom.mjs and dist/amp/worker/worker.mjs

erwinmombay:
Since we distribute these files as the module version ideally they should be optimized from the get go instead of optimizing them on the user side (such as the AMP library)

### DIFF
--- a/config/rollup.worker-thread.js
+++ b/config/rollup.worker-thread.js
@@ -107,6 +107,7 @@ const ESModules = [
         transpileToES5: false,
         allowConsole: true,
       }),
+      ...compilePlugins,
     ],
   },
   {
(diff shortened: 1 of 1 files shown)


erwinmombay:
once this new code is published, i'll update the version we're pulling from in the https://github.com/ampproject/amphtml as well as update the copy/dist code so that it now copies the correct files instead of turning the "mjs" files into ".max.js" files

erwinmombay:
@samouri @kristoferbaxter PTAL. im not able to assign reviewers (do i need to apply for collaborator access?)

kristoferbaxter:
https://github.com/ampproject/worker-dom/pull/882#issuecomment-661322037

We should revisit this though, it's confusing.

erwinmombay:
this PR is super-ceded by https://github.com/ampproject/worker-dom/pull/929

PR #929, opened 2020-09-03
Refactor output

kristoferbaxter:
Move to `amp-debug` and `amp-production` outputs.


config/rollup.main-thread.js
```diff
@@ -62,13 +62,36 @@ const ESModules = [
...
+      }),
+      babelPlugin({
+        transpileToES5: false,
+        allowConsole: false,
+      }),
+      compiler(),
+      terser(),
+    ],
+  },
+  {
```
samouri, line 83: note: we don't currently have the setup to be able to switch to main-thread debug variant ([1](https://github.com/ampproject/amphtml/blob/master/extensions/amp-script/0.1/amp-script.js#L17)), but this would be nice to add.

config/rollup.worker-thread.js
```diff
@@ -71,11 +71,29 @@ const ESModules = [
...
+    ],
+  },
+  {
+    input: 'output/worker-thread/index.amp.js',
+    output: {
+      file: 'dist/amp-debug/worker/worker.mjs',
       format: 'iife',
       name: 'WorkerThread',
       sourcemap: true,
-      banner: 'var WORKER_DOM_DEBUG = /log|development/i.test(location.hash);',
```
samouri, line 78: happy to see this go away :). We'll need to pick the correct one based on dev/prod builds.

samouri (changes requested):
LGTM. 
We also need to update some of the filesize checks (and add the ones that don't exist yet. This is something I had missed when adding `nodom`)

kristoferbaxter:
Added tracking for the `nodom` variant.

samouri (approved):

PR #937, opened 2020-09-08
Output MJS only, requiring downstream consumers to transpile if necessary

kristoferbaxter:
Not sure this is a good idea, but wanted to share the working version of this.


erwinmombay:
@kristoferbaxter would you mind sharing your thoughts on why this is would be a good idea? (picking your brain for the pros and cons here)
is this to have downstream have more control on the transpiled code? or is this to kind of to make downstream have to modernize? or more on the maintenance cost on our end?

kristoferbaxter:
Pros:
- Library is simpler to build and only outputs valid latest syntax.
- Can move to faster compilation systems (like esbuild).

Cons:
- Consumers have to transpile to their own settings.
- Minification will likely be worse (as application minification isn't tweaked to the same degree as this libraries)
- Potential maintenance burdon (this library requires strict transpilation, loose mode fails silently)

I lean toward *not* doing this change, but wanted to share the results.

samouri:
I'd also lean against this change for now,  due to:
> Consumers have to transpile to their own settings.

kristoferbaxter:
Closing based on discussion. Thanks everyone!

Issue #613, opened 2019-08-15
HTML entities added to innerHTML are not decoded


westonruter:
Consider this `amp-script`:

```html
  <amp-script script="entity-test">
    <p>
      Loading...
    </p>
  </amp-script>
  <script id="entity-test" type="text/plain" target="amp-script">
  	document.querySelector('p').innerHTML = '&lsaquo;&#8249;&#x2039; &#x203A;&#8250;&rsaquo;';
  </script>

  <!-- Invalid AMP for debugging difference in innerHTML handling -->
  <button type=button onclick="eval( document.getElementById( 'entity-test' ).textContent )">
    Try running script in main thread
  </button>
```

The rendered output is:

```
&lsaquo;&#8249;&#x2039; &#x203A;&#8250;&rsaquo;
```

When I expected the HTML entities to be decoded as:

```html
‹‹‹ ›››
```

So WorkerDOM is not properly decoding HTML entities.

kristoferbaxter:
Am investigating solutions here given there is no JS API I'm aware of that decodes html entities.

jridgewell:
The label based ones need a map, the integers can call `String.fromCharCode`.

westonruter:
Yes, I was just going to say: Decoding of numeric (decimal and hexadecimal) entities seems straightforward. Named entities, however, will need a lookup table.

samouri:
The HTML spec website has a list of all the html entities: https://html.spec.whatwg.org/entities.json

jridgewell:
Yup, but that table would be 10kb brotli'd, even after we remove the `codepoints` key (which we wouldn't need).

samouri:
What do folks think about exposing a new function, `innerHTMLAsync`. It would directly send over the text to the main-thread to apply via `innerHTML`, and then worker-dom would refresh the vdom for the relevant node (delete/rehydrate)

samouri:
I've created a PR that just decodes numeric entities (https://github.com/ampproject/worker-dom/pull/947). I think it could also be worth decoding the 4 named entities required to encode html, since they are by far the most common (< >, &, ").

PR #947, opened 2020-09-17
decode numeric html entitites

samouri:
Addressed the easy half of https://github.com/ampproject/worker-dom/issues/613

cc @caroqliu.

Note that we can get a really big bang for the innerHTML buck if we also decide to decode just the 4 required bits for [html encoding](https://developer.mozilla.org/en-US/docs/Glossary/Entity) (`<` `>`, `&`, `"`). Could be worth doing in addition to all numeric.

### DIFF
--- a/src/third_party/html-parser/html-parser.ts
+++ b/src/third_party/html-parser/html-parser.ts
@@ -2,6 +2,7 @@ import { Element } from '../../worker-thread/dom/Element';
 import { Node } from '../../worker-thread/dom/Node';
 import { SVG_NAMESPACE, HTML_NAMESPACE } from '../../transfer/TransferrableNodes';
 import { toLower, toUpper } from '../../utils';
+import { CharacterData } from '../../worker-thread/dom/CharacterData';
 
 interface Elements {
   [key: string]: boolean;
@@ -235,7 +236,7 @@ export function parse(data: string, rootElement: Element) {
 
 function decodeNumericEntities(html: string) {
   return html.replace(/&#(x?[\da-f]+);?/gi, function (s, entity) {
-    let code = entity.charAt(0) === 'x' ? parseInt(entity.substr(1).toLowerCase(), 16) : parseInt(entity, 10);
+    let code = entity.charAt(0).toLowerCase() === 'x' ? parseInt(entity.substr(1).toLowerCase(), 16) : parseInt(entity, 10);
 
     // 1114111 is the largest valid unicode codepoint.
     if (isNaN(code) || code > 1114111) {
--- a/src/third_party/html-parser/html-parser.ts
+++ b/src/third_party/html-parser/html-parser.ts
@@ -128,7 +128,7 @@ export function parse(data: string, rootElement: Element) {
     if (lastTextPos < match.index) {
       // if has content
       const text = data.slice(lastTextPos, match.index);
-      currentParent.appendChild(ownerDocument.createTextNode(text));
+      currentParent.appendChild(ownerDocument.createTextNode(decodeNumericEntities(text)));
     }
     lastTextPos = kMarkupPattern.lastIndex;
     if (commentContents !== undefined) {
@@ -231,4 +231,16 @@ export function parse(data: string, rootElement: Element) {
   }
 
   throw new Error('Attempting to parse invalid HTML.');
+}
+
+function decodeNumericEntities(html: string) {
+  return html.replace(/&#(x?\d+);?/g, function (s, entity) {
+    const code = entity.charAt(0) === 'x' ? parseInt(entity.substr(1).toLowerCase(), 16) : parseInt(entity, 10);
+
+    let chr;
+    if (!(isNaN(code) || code < -32768 || code > 65535)) {
+      chr = String.fromCharCode(code);
+    }
+    return chr || s;
+  });
 }
--- a/src/third_party/html-parser/html-parser.ts
+++ b/src/third_party/html-parser/html-parser.ts
@@ -234,11 +234,11 @@ export function parse(data: string, rootElement: Element) {
 }
 
 function decodeNumericEntities(html: string) {
-  return html.replace(/&#(x?\d+);?/g, function (s, entity) {
+  return html.replace(/&#(x?[\da-zA-Z]+);?/g, function (s, entity) {
     const code = entity.charAt(0) === 'x' ? parseInt(entity.substr(1).toLowerCase(), 16) : parseInt(entity, 10);
 
     let chr;
-    if (!(isNaN(code) || code < -32768 || code > 65535)) {
+    if (!(isNaN(code) || code < 0 || code > 65535)) {
       chr = String.fromCharCode(code);
     }
     return chr || s;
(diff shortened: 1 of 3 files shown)


src/third_party/html-parser/html-parser.ts
```diff
@@ -232,3 +232,15 @@ export function parse(data: string, rootElement: Element) {
 
   throw new Error('Attempting to parse invalid HTML.');
 }
+
+function decodeNumericEntities(html: string) {
+  return html.replace(/&#(x?\d+);?/g, function (s, entity) {
```
jridgewell, line 237: We need to parse hex entities and number entities.
```suggestion
  return html.replace(/&#(?:x([a-f\d]+)|(\d+));?/ig, function (s, hex, num) {
```

src/third_party/html-parser/html-parser.ts
```diff
@@ -232,3 +232,15 @@ export function parse(data: string, rootElement: Element) {
...
+    const code = entity.charAt(0) === 'x' ? parseInt(entity.substr(1).toLowerCase(), 16) : parseInt(entity, 10);
+
+    let chr;
+    if (!(isNaN(code) || code < -32768 || code > 65535)) {
```
jridgewell, line 241: What's up with -32768? But entities can definitely go above 65535, eg `&#x1F4A9;`

jridgewell (changes requested):

src/third_party/html-parser/html-parser.ts
samouri: ah right, hex means letters too

src/third_party/html-parser/html-parser.ts
samouri: Good choice of example :P. So that entity is actually parsed as multiple codepoints. See the mdn description: 
```
A sequence of numbers that are UTF-16 code units. The range is between 0 and 65535 (0xFFFF). Numbers greater than 0xFFFF are truncated. No validity checks are performed.
```

`&#x1F4A9;` --> `String.fromCharCode([55357, 56489])`

src/third_party/html-parser/html-parser.ts
```diff
@@ -232,3 +232,15 @@ export function parse(data: string, rootElement: Element) {
...
   throw new Error('Attempting to parse invalid HTML.');
 }
+
+function decodeNumericEntities(html: string) {
+  return html.replace(/&#(x?[\da-zA-Z]+);?/g, function (s, entity) {
+    const code = entity.charAt(0) === 'x' ? parseInt(entity.substr(1).toLowerCase(), 16) : parseInt(entity, 10);
+
+    let chr;
+    if (!(isNaN(code) || code < 0 || code > 65535)) {
+      chr = String.fromCharCode(code);
```
jridgewell, line 242: We still need to match 💩. When greater than 65535, you can calculate the code points as:

```js
function encode(code) {
  code -= 0x10000;
  return [0xd800 | ((code >> 10) & 0x3ff), 0xdc00 | (code & 0x3ff)];
}
String.fromCharCode(...encode(0x1F4A9))
```

src/third_party/html-parser/html-parser.ts
```diff
@@ -232,3 +232,15 @@ export function parse(data: string, rootElement: Element) {
 
   throw new Error('Attempting to parse invalid HTML.');
 }
+
+function decodeNumericEntities(html: string) {
+  return html.replace(/&#(x?[\da-zA-Z]+);?/g, function (s, entity) {
+    const code = entity.charAt(0) === 'x' ? parseInt(entity.substr(1).toLowerCase(), 16) : parseInt(entity, 10);
+
+    let chr;
+    if (!(isNaN(code) || code < 0 || code > 65535)) {
```
jridgewell, line 241: From our regex, it's impossible to get a negative number.

src/third_party/html-parser/html-parser.ts
```diff
@@ -232,3 +232,15 @@ export function parse(data: string, rootElement: Element) {
 
   throw new Error('Attempting to parse invalid HTML.');
 }
+
+function decodeNumericEntities(html: string) {
+  return html.replace(/&#(x?[\da-zA-Z]+);?/g, function (s, entity) {
```
jridgewell, line 237: Hex is `a-f`, not `a-z`.

src/third_party/html-parser/html-parser.ts
samouri: can you...explain how this works?

samouri: Ah, ok. Just read about surrogate pairs, so there can only ever be 2. Any objection to using `fromCodePoint` instead?

patrickkettner: I don't believe we were ever able to ship `fromCodePoint` in IE11, if you are needing to support that

samouri: IE11 support is already broken until we [start polyfilling](https://github.com/ampproject/worker-dom/pull/886/files). Once we are, `fromCodePoint` has a [polyfill available](https://github.com/zloirock/core-js/blob/v3.6.5/packages/core-js/modules/es.string.from-code-point.js) in core-js.

src/third_party/html-parser/html-parser.ts
```diff
@@ -232,3 +232,16 @@ export function parse(data: string, rootElement: Element) {
 
   throw new Error('Attempting to parse invalid HTML.');
 }
+
+function decodeNumericEntities(html: string) {
+  return html.replace(/&#(x?[\da-fA-F]+);?/g, function (s, entity) {
```
jridgewell, line 237: The `x` can be `X`, too.

```suggestion
  return html.replace(/&#(x?[\da-f]+);?/gi, function (s, entity) {
```

jridgewell (approved):

src/third_party/html-parser/html-parser.ts
```diff
@@ -234,7 +234,7 @@ export function parse(data: string, rootElement: Element) {
 }
 
 function decodeNumericEntities(html: string) {
-  return html.replace(/&#(x?[\da-fA-F]+);?/g, function (s, entity) {
+  return html.replace(/&#(x?[\da-f]+);?/gi, function (s, entity) {
     let code = entity.charAt(0) === 'x' ? parseInt(entity.substr(1).toLowerCase(), 16) : parseInt(entity, 10);
```
jridgewell, line 238: This needs to be updated now. That's why I had split them in the initial https://github.com/ampproject/worker-dom/pull/947#discussion_r490598164

samouri: you mean to accept a capital `X` right? Fixed.

src/third_party/html-parser/html-parser.ts
```diff
@@ -2,6 +2,7 @@ import { Element } from '../../worker-thread/dom/Element';
 import { Node } from '../../worker-thread/dom/Node';
 import { SVG_NAMESPACE, HTML_NAMESPACE } from '../../transfer/TransferrableNodes';
 import { toLower, toUpper } from '../../utils';
+import { CharacterData } from '../../worker-thread/dom/CharacterData';
```
jridgewell: Unintentional import?

jridgewell (approved):

src/third_party/html-parser/html-parser.ts
samouri: yeah, need to turn off auto-importing in vs code. will remove now?

samouri: hmm....already did actually

kristoferbaxter (approved):

PR #948, opened 2020-09-17
implement setInnerHTMLAsync

samouri:
**summary**

Creates a new function on `Element` called `setInnerHTMLAsync(htmlStr)`. It functions exactly like `innerHTML = htmlStr` would on the main thread. This is because the implementation passes the string directly along to the main thread which actually sets innerHTML, and then the vdom is rehydrated.


samouri:
Closing since there is no current need for this. Can reopen in the future if that changes.

PR #952, opened 2020-09-18
Decode HTML Reserved Characters (>,<,&,").

samouri:
Followup to https://github.com/ampproject/worker-dom/pull/947. I think it may make sense to decode these 4 in particular, given their prominence in html encoding. See https://developer.mozilla.org/en-US/docs/Glossary/Entity

### DIFF
--- a/src/third_party/html-parser/html-parser.ts
+++ b/src/third_party/html-parser/html-parser.ts
@@ -2,6 +2,7 @@ import { Element } from '../../worker-thread/dom/Element';
 import { Node } from '../../worker-thread/dom/Node';
 import { SVG_NAMESPACE, HTML_NAMESPACE } from '../../transfer/TransferrableNodes';
 import { toLower, toUpper } from '../../utils';
+import { CharacterData } from '../../worker-thread/dom/CharacterData';
 
 interface Elements {
   [key: string]: boolean;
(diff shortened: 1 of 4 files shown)


src/third_party/html-parser/html-parser.ts
```diff
@@ -232,3 +232,31 @@ export function parse(data: string, rootElement: Element) {
...
+
+/**
+ * Decodes HTML Entities.
+ * Currently only works for numeric entities, as well as the 4 named entities
+ * required for encoding HTML: &, ", <, >.
+ * See https://developer.mozilla.org/en-US/docs/Glossary/Entity.
+ *
+ * TODO: create solution for other named entities.
+ */
+const RESERVED_CHARACTERS: { [key: string]: string } = { amp: '&', lt: '<', gt: '>', quot: '"' };
```
jridgewell, line 244: Since we’re doing unchecked property lookups, make this a prototype-less object.

samouri: How? is that the `Object.create(null)` trick?

kristoferbaxter (approved):

src/third_party/html-parser/html-parser.ts
jridgewell: Yah. Or, `{ __proto__: null, amp: '&', … }` which is a literal syntax equivalent.

src/third_party/html-parser/html-parser.ts
```diff
@@ -233,15 +233,30 @@ export function parse(data: string, rootElement: Element) {
...
 
-    // 1114111 is the largest valid unicode codepoint.
-    if (isNaN(code) || code > 1114111) {
-      return s;
+      return String.fromCodePoint(code) || s;
     }
 
-    return String.fromCodePoint(code) || s;
+    // Named entity. Only HTML reserved entities are currently supported.
+    return RESERVED_CHARACTERS[namedEntity] || s;
```
jridgewell, line 260: Oh, I think we need to lowercase, too?

src/third_party/html-parser/html-parser.ts
samouri: can't do that with my current typing :/

src/third_party/html-parser/html-parser.ts
samouri: yeah, good call. Done

src/third_party/html-parser/html-parser.ts
samouri: updated the typing to allow for the null value. done

jridgewell (approved):

PR #989, opened 2020-10-28
Fix race between long task start and end handlers.

samouri:
**problem**
Consumers of `worker-dom` using the `longTask` API are notified of a new long task via callback and are notified of a long task end via Promise. In general this is fine.

There are situation in which both of these events occur within the same task. This is particularly true when the page is inactive and therefore `postMessage` processing is delayed until the page has been returned to.

A naive consumer will then run into the strange situation where they are notified of a 2nd long task start before receiving the 1st end. For `<amp-script>` this meant a sequence of events like this:

1. Long task start
2. Long task start
3. Long task end  (now `inLongTask=false`)
4. `<amp-script>` mutation pump receives mutations but inLongTask=false and activation time was set by (3), and therefore it panics.

**solution notes**
Initially I had solved this in `<amp-script>` via counter, but that felt wasteful. Instead I think its more elegant to sequence the `config.longTask` call after a `Promise.resolve()`, which would be guaranteed to run after the consumers `.then()`.

Note that all of the pre-existing tests I've modified would have passed with the added `await sleep(0)`s  even before this PR, whereas the newly introduced test would have failed.

Fixes: https://github.com/ampproject/amphtml/issues/30548

### DIFF
--- a/src/main-thread/commands/long-task.ts
+++ b/src/main-thread/commands/long-task.ts
@@ -53,6 +53,12 @@ export const LongTaskExecutor: LongTaskCommandExecutorInterface = (
           index++;
           if (!currentResolver) {
             const newResolver = new Promise((resolve) => (currentResolver = resolve));
+            // One of the worker-dom contracts is that there should not be two
+            // LONG_TASK_STARTs in a row without an END in between. In case both exist within
+            // the same set of mutations, a naive consumer will likely have a race since START is invoked
+            // synchronously via callback, whereas END must be chained to a resolving promise and therefore
+            // would occur afterwards.
+            // We guard against this by yielding to the event loop before calling longTask.
             setTimeout(() => {
               config.longTask?.(newResolver);
             }, 0);
--- a/src/main-thread/commands/long-task.ts
+++ b/src/main-thread/commands/long-task.ts
@@ -55,13 +55,15 @@ export const LongTaskExecutor: LongTaskCommandExecutorInterface = (
             const newResolver = new Promise((resolve) => (currentResolver = resolve));
             // One of the worker-dom contracts is that there should not be two
             // LONG_TASK_STARTs in a row without an END in between. In case both exist within
-            // the same set of mutations, a naive consumer will likely have a race since START is invoked
-            // synchronously via callback, whereas END must be chained to a resolving promise and therefore
-            // would occur afterwards.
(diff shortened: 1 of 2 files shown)


src/main-thread/commands/long-task.ts
```diff
@@ -52,7 +52,18 @@ export const LongTaskExecutor: LongTaskCommandExecutorInterface = (
...
           if (!currentResolver) {
-            config.longTask(new Promise((resolve) => (currentResolver = resolve)));
+            const newResolver = new Promise((resolve) => (currentResolver = resolve));
+            // One of the worker-dom contracts is that there should not be two
+            // LONG_TASK_STARTs in a row without an END in between. In case both exist within
+            // the same set of mutations, we need to guard against having a consumers 1st END
+            // handler occur after the START handler. If we synchronously called longTask() here it
+            // would likely occur due to scheduling of callbacks vs. promise.
+            // See: worker-dom/pull/989/files
+            Promise.resolve().then(() => {
```
kristoferbaxter, line 62: Does the return type matter here?

`Promise.resolve().then(() => config.longTask ? config.longTask(newResolver));`

src/test/main-thread/long-task.test.ts
```diff
@@ -21,9 +21,13 @@ import { TransferrableMutationType } from '../../transfer/TransferrableMutation'
 import { StringContext } from '../../main-thread/strings';
 import { NodeContext } from '../../main-thread/nodes';
 import { WorkerContext } from '../../main-thread/worker';
-import { normalizeConfiguration } from '../../main-thread/configuration';
+import { normalizeConfiguration, WorkerDOMConfiguration } from '../../main-thread/configuration';
 import { ObjectContext } from '../../main-thread/object-context';
 
+async function sleep(ms: number) {
```
kristoferbaxter, line 27: Nit: Can you also indicate the return type?

`function sleep(ms: number): Promise<...>`

kristoferbaxter (approved):
New tests look great, and the explanation is clear.

Nice work! And thank you.

src/main-thread/commands/long-task.ts
samouri: Not at all. Updating

src/test/main-thread/long-task.test.ts
samouri: whoops, adding now

PR #991, opened 2020-10-29
Iframify

samouri:
**summary**
This is not meant for merging. It is an exploration / prototype to see what it would take to switch to a sandboxed iframe.

- We need  to message back to main thread when the iframe was initialized and delay sending messages from main-thread to iframe until that point.
- For `yesdom` (standard) mode,  we would need to make sure the vdom classes installed by worker-dom don't clobber with the real iframe provided ones. (remove ones provided by iframe first?)
- The messaging channels seem to work with only slight tweaking (postMessage, vs. iframe.contentWindow.postMessage).

Questions to answer:
- why did we originally pick Worker vs. Iframe. Do the reasons still hold up? What are the security implications of one vs. the other (are we sure Worker isn't secure enough for the use-case at hand)
- Does it need to be a Worker within an iframe, or is iframe enough?

Note: it is also possible to use standard WorkerDOM within an iframe and do it all within amp-script, but I can't see a way of doing that without also forcing the client to download worker-dom main-thread code twice.

### DIFF
--- a/config/dist-packaging/amp/package.json
+++ b/config/dist-packaging/amp/package.json
@@ -1,0 +1,10 @@
+{
+  "main": "main.mjs",
+  "browser": {
+    "./main": "main.mjs",
+    "./worker": "worker.mjs"
+  },
+  "syntax": {
+    "esmodules": "main.mjs"
+  }
+}
--- a/config/dist-packaging/amp/worker/package.json
+++ b/config/dist-packaging/amp/worker/package.json
@@ -1,0 +1,6 @@
+{
+  "main": "worker.mjs",
+  "syntax": {
+    "esmodules": "worker.mjs"
+  }
+}
--- a/config/dist-packaging/debug/package.json
+++ b/config/dist-packaging/debug/package.json
@@ -1,0 +1,10 @@
+{
+  "main": "main.js",
+  "browser": {
+    "./main": "main.js",
+    "./worker": "worker.js"
+  },
+  "syntax": {
+    "esmodules": "main.mjs"
+  }
+}
--- a/config/dist-packaging/debug/worker/package.json
+++ b/config/dist-packaging/debug/worker/package.json
@@ -1,0 +1,6 @@
+{
+  "main": "worker.js",
+  "syntax": {
+    "esmodules": "worker.mjs"
+  }
+}
--- a/config/dist-packaging/worker/package.json
+++ b/config/dist-packaging/worker/package.json
@@ -1,0 +1,6 @@
+{
+  "main": "worker.js",
+  "syntax": {
+    "esmodules": "worker.mjs"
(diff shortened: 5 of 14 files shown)


samouri:
Just learned that iframe js run in the same thread as the main doc, so there is still a major perf reason to run within a Worker. This means the optimal setup is likely `<amp-script>` asking worker-dom to create an iframed workerdom. The `worker-dom` main-thread code still must run within the the main doc, and then instead of creating a regular worker, it creates an iframe with a `srcdoc` that creates a worker as well as proxies all messages. This also evades all the problems of trying to make worker-dom behave well within an iframe instead of Worker environment (globals clashing)

zhouyx:
>The worker-dom main-thread code still must run within the the main doc, and then instead of creating a regular worker, it creates an iframe with a srcdoc that creates a worker as well as proxies all messages. This also evades all the problems of trying to make worker-dom behave well within an iframe instead of Worker environment (globals clashing)

Agreed. But I can't create non origin iframe with srcdoc. The reason being worker within non origin iframe runs under the same origin of the parent doc in this case. This was not an issue with `<amp-script>` before because of the script hash requirement. But it will become an issue if we were to remove it. 

I have a proof of concept PR that wrap the worker inside a cross origin iframe. https://github.com/ampproject/amphtml/pull/30920/files#r515393488

samouri:
> The reason being worker within non origin iframe runs under the same origin of the parent doc in this case. 

Is this still important even with a full `sandbox="allow-scripts"`?

> I have a proof of concept PR that wrap the worker inside a cross origin iframe.

The issue with doing it in amphtml IMO is that the iframe will need to

zhouyx:
Unfortunately yes. I was also very surprised that the `sandbox` doesn't provide the isolation to worker inside it. 
 [Here's](https://glitch.com/edit/#!/safe-worker?path=test-indexdb-blob.html%3A16%3A41) how I tested it.

samouri:
Instead of `srcdoc`, could we use a `data:text/html` src? for example:

```js
function createIframe(html) {
 const iframe = document.createElement('iframe');
 iframe.src = "data:text/html," + html;
 iframe.setAttribute('sandbox', 'allow-scripts');
 iframe.setAttribute('style', 'display:none');
 return iframe;
}
```

I just tested on Chrome and it creates a "null origin" iframe

zhouyx:
I'm not familiar with `data:text/html` src. Worth trying. 
But my guess is that non origin iframe won't help much here. srcdoc also creates a non origin iframe, but the worker script loaded using blobUrl was from the parent doc's origin, not non origin. 

I can modify my test page and try this later today.

samouri:
Fully implemented in https://github.com/ampproject/worker-dom/pull/1042

PR #1006, opened 2020-11-17
storage should throw if unsupported

samouri:
**summary**
Fixes https://github.com/ampproject/amphtml/issues/30691. Makes it so that worker-dom no longer blows up when storage access is denied by the browser.  From the worker's perspective, localStorage and sessionStorage end up as null.

- [x] Write tests.
- [ ] Need to investigate if sanitizer storage init is borked. My impression is that it serializes a promise, which I don't think does what we want. Can pursue this separately though, since if I'm right then it was already broken.

### DIFF
--- a/demo/iframe/index.html
+++ b/demo/iframe/index.html
@@ -1,11 +1,12 @@
-<!-- 
-  This file added to assist in testing WorkerDOM iframed use cases. 
+<!--
+  This file added to assist in testing WorkerDOM iframed use cases.
   Notably, AMP Documents are often displayed in an crossorigin iframe.
   These documents have various restrictions, including no access to storage from
   within incognito windows.
 
-  In order to run this demo, you need to map workerdom.localhost to 
-  127.0.0.1 in your /etc/hosts file.
+  To run this demo:
+    - you need to map workerdom.localhost to 127.0.0.1 in your /etc/hosts file.
+    - load this page in both regular and incognito modes.
  -->
 <html>
   <head></head>
--- a/demo/iframe/index.html
+++ b/demo/iframe/index.html
@@ -1,0 +1,6 @@
+<html>
+  <head></head>
+  <body>
+    <iframe src="http://workerdom.localhost:3001/demo/iframe/iframe.html"></iframe>
+  </body>
+</html>
--- a/demo/iframe/index.html
+++ b/demo/iframe/index.html
@@ -1,3 +1,12 @@
+<!-- 
+  This file added to assist in testing WorkerDOM iframed use cases. 
+  Notably, AMP Documents are often displayed in an crossorigin iframe.
+  These documents have various restrictions, including no access to storage from
+  within incognito windows.
+
+  In order to run this demo, you need to map workerdom.localhost to 
+  127.0.0.1 in your /etc/hosts file.
+ -->
 <html>
   <head></head>
   <body>
--- a/demo/iframe/index.html
+++ b/demo/iframe/index.html
@@ -11,6 +11,6 @@
 <html>
   <head></head>
   <body>
-    <iframe src="http://workerdom.localhost:3001/demo/iframe/iframe.html"></iframe>
+    <iframe src="http://localhost:3001/demo/iframe/iframe.html"></iframe>
   </body>
 </html>
--- a/demo/iframe/storage.js
+++ b/demo/iframe/storage.js
@@ -1,0 +1,12 @@
+try {
+  localStorage.setItem('marco', 'polo');
+  console.log('localStorage is available');
+} catch (e) {
+  console.error(e);
+}
+try {
+  sessionStorage.setItem('marco', 'polo');
+  console.log('sessionStorage is available');
+} catch (e) {
+  console.error(e);
+}
--- a/demo/iframe/storage.js
+++ b/demo/iframe/storage.js
@@ -1,12 +1,14 @@
 try {
   localStorage.setItem('marco', 'polo');
-  console.log('localStorage is available');
+  console.log('localStorage is available at: ' + window.location);
 } catch (e) {
   console.error(e);
 }
 try {
   sessionStorage.setItem('marco', 'polo');
-  console.log('sessionStorage is available');
+  console.log('sessionStorage is available at: ' + window.location);
 } catch (e) {
   console.error(e);
 }
+
+console.log('Can run code after the storage access occurs');
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
       "brotli": "5 kB"
     },
     "./dist/worker/worker.mjs": {
-      "brotli": "11.85 kB"
+      "brotli": "12 kB"
     },
     "./dist/worker/worker.js": {
       "brotli": "13.5 kB"
--- a/package.json
+++ b/package.json
@@ -96,7 +96,7 @@
       "brotli": "5 kB"
     },
     "./dist/worker/worker.mjs": {
-      "brotli": "11.76 kB"
+      "brotli": "11.85 kB"
     },
     "./dist/worker/worker.js": {
       "brotli": "13.5 kB"
--- a/src/main-thread/worker.ts
+++ b/src/main-thread/worker.ts
@@ -22,7 +22,9 @@ import { NodeContext } from './nodes';
 import { TransferrableKeys } from '../transfer/TransferrableKeys';
 import { StorageLocation } from '../transfer/TransferrableStorage';
 
-export type StorageSupport = { supported: boolean; errorMsg?: string };
+// TODO: Sanitizer storage init is likely broken, since the code currently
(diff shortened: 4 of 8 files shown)


src/main-thread/worker.ts
```diff
@@ -67,11 +79,18 @@ export class WorkerContext {
...
           [${window.innerWidth}, ${window.innerHeight}],
-          ${JSON.stringify(localStorageData)},
-          ${JSON.stringify(sessionStorageData)}
+          ${JSON.stringify(localStorageSupportedStatus.errorMsg ?? localStorageData)},
+          ${JSON.stringify(sessionStorageSupportedStatus.errorMsg ?? sessionStorageData)}
         );
         workerDOM.document[${TransferrableKeys.observe}](this);
-        Object.keys(workerDOM).forEach(function(k){self[k]=workerDOM[k]});
+        Object.keys(workerDOM).forEach(function(k){
+          /* These two are already assigned to Window. They also have the potential to throw
```
samouri, line 87: self nit: this comment will end up in the output. move to a real comment.

src/main-thread/worker.ts
```diff
@@ -104,3 +123,13 @@ export class WorkerContext {
     this.worker.postMessage(message, transferables || []);
   }
 }
+
+function getStorageSupportStatus(type: 'localStorage' | 'sessionStorage'): StorageSupport {
+  try {
+    let randKey = 'WORKER_DOM_TEST_KEY';
+    globalThis[type].setItem(randKey, '');
```
samouri, line 130: nit: this is main thread, so actually use `window`

demo/iframe/index.html
```diff
@@ -0,0 +1,16 @@
+<!--
+  This file added to assist in testing WorkerDOM iframed use cases.
+  Notably, AMP Documents are often displayed in an crossorigin iframe.
+  These documents have various restrictions, including no access to storage from
+  within incognito windows.
+
+  To run this demo:
+    - you need to map workerdom.localhost to 127.0.0.1 in your /etc/hosts file.
```
kristoferbaxter, line 8: Nit: Can this file instead be a custom response using two ports so we don't need an /etc/hosts file override? This would allow us to down the line use these 'demos' as e2e test environments.

demo/iframe/storage.js
```diff
@@ -0,0 +1,12 @@
+try {
+  localStorage.setItem('marco', 'polo');
+  console.log('localStorage is available');
+} catch (e) {
+  console.error(e);
+}
+try {
+  sessionStorage.setItem('marco', 'polo');
```
kristoferbaxter, line 8: Hah, marco polo is quite a funny example here.

package.json
```diff
@@ -96,7 +96,7 @@
       "brotli": "5 kB"
     },
     "./dist/worker/worker.mjs": {
-      "brotli": "11.76 kB"
+      "brotli": "11.85 kB"
```
kristoferbaxter, line 99: Let's bump this to 12 perhaps to give a bit more room? `11.76` was a strange value.

src/main-thread/worker.ts
```diff
@@ -67,11 +72,13 @@ export class WorkerContext {
...
           ${JSON.stringify(globalEventHandlerKeys)},
           [${window.innerWidth}, ${window.innerHeight}],
-          ${JSON.stringify(localStorageData)},
-          ${JSON.stringify(sessionStorageData)}
+          ${JSON.stringify(localStorageInit)},
+          ${JSON.stringify(sessionStorageInit)}
         );
         workerDOM.document[${TransferrableKeys.observe}](this);
-        Object.keys(workerDOM).forEach(function(k){self[k]=workerDOM[k]});
+        Object.keys(workerDOM).forEach(function(k){
```
kristoferbaxter, line 79: Nit: Can we validate this code is run thru a minifier? Otherwise we should remove the newline.

src/test/main-thread/initialize-storage.test.ts
```diff
@@ -0,0 +1,64 @@
...
+  const unsupportedAccess: WorkerStorageInit = {
+    supported: false,
+    errorMsg: 'Access denied',
+  };
+  const { doc, self } = t.context;
+  initialize(doc, [], {} as HydrateableNode, [], [], [0, 0], unsupportedAccess, unsupportedAccess);
+  t.throws(() => doc.defaultView.localStorage, {
+    message: unsupportedAccess.errorMsg,
+  });
+  t.throws(() => self.localStorage, { message: unsupportedAccess.errorMsg });
```
kristoferbaxter, line 43: Thanks for adding tests here.

src/worker-thread/initialize.ts
```diff
@@ -42,6 +45,31 @@ export function initialize(
...
+  if (sessionStorageInit.supported) {
+    window.sessionStorage = createStorage(document, StorageLocation.Session, sessionStorageInit.storage);
+    (self as any).sessionStorage = window.sessionStorage;
+  } else {
+    defineThrowingGlobal(window, 'sessionStorage', sessionStorageInit.errorMsg);
+  }
+}
+
+function defineThrowingGlobal(win: WorkerDOMGlobalScope, property: string, errorMsg: string) {
+  Object.defineProperty(win, property, {
```
kristoferbaxter, line 63: Should the value of the Object just be null or is throwing required?

kristoferbaxter (commented):
Left a few comments and nits, but overall looks great.

I'll approve after we chat about the few comments.

package.json
samouri: that was actually my first instinct as well! I was worried you would comment to push it back down hah.
I'll push to 12

src/worker-thread/initialize.ts
samouri: If we want to match browser, then we need to throw

demo/iframe/index.html
samouri: Discussed offline. Not possible right now since different ports in the same origin seem to be treated as same origin by the check

samouri: Even more updates. Turns out localhost appears to be whitelisted, and the only reason crossorigin simulation was working was because Chrome wasn't smart enough to realize the `/etc/hosts/` mapped domain was actually also pointed to localhost.

Safari and FF both got this correct. So assuming Chrome eventually gets smarter here, then this demo would be useless on all browsers.

The cleanest solution is to instead manually redefine localStorage and sessionsStorage to throw when accessed (simulating a crossorigin iframe experience) instead.

Updated / WDYT?

src/main-thread/worker.ts
samouri: by choosing null instead of throwing, this piece was reverted

src/worker-thread/initialize.ts
samouri: decided not to match browser

src/worker-thread/initialize.ts
```diff
@@ -23,15 +23,17 @@ import { createStorage } from './Storage';
 import { StorageLocation } from '../transfer/TransferrableStorage';
 import { appendGlobalEventProperties } from './dom/HTMLElement';
 
+export type WorkerStorageInit = { supported: true; storage: { [key: string]: string } } | { supported: false; errorMsg: string };
```
kristoferbaxter, line 26: This does mean the return type is different for supported or not... could we simplify this?

```suggestion
export type WorkerStorageInit = {
  storage: null | { [key: string]: string }
  error: null | string;
}
```

If storage is not `null` use the value, otherwise output the `error`.

samouri: From a domain modeling perspective, that would imply that it is possible for `{storage: null, error: null}` object to exist. When possible, I'd prefer to make incorrect states impossible to even represent. 

I do see your point though that `success` can be implicit and derived from the existence of storage or error. I'm not sure how to accomplish that without providing a shared key to discriminate on.

kristoferbaxter: I'm mostly concerned about the shape of the return object being different given inputs.

Another idea:

```suggestion
export type WorkerStorageInit = { 
  storage: { [key: string]: string };
  error: null;
} | {
  storage: null;
  error: string;
};
```

samouri: Ah! figured it out with the `in` operator for type narrowing. Pushing an update now

samouri: >I'm mostly concerned about the shape of the return object being different given inputs.

For v8 reasons? I like your latest shape, updating now

kristoferbaxter: > For v8 reasons? I like your latest shape, updating now

Yep, it's super minor here but in general keeping the same shape is preferred for a method return value (afaik).

kristoferbaxter (approved):
Last item to address is if we want to `console.warn` when storage isn't available.

Likely ok to do though we might reconsider if spammy.

PR #1016, opened 2021-01-06
stub worker wrapped in iframe

zhouyx:


samouri:
Closing as this FR was implemented in https://github.com/ampproject/worker-dom/pull/1042

Issue #1023, opened 2021-02-24
Possible to support childNode.replaceWith and childNode.remove?


Slapbox:
I've been reviewing the capabilities and it's really impressive!

I was hoping we'd be able to switch over to using worker-dom, but I found as I got to the bottom of our code, the last two methods we use are not supported.

Is it possible to add support for `childNode.replaceWith` and `childNode.remove`? 

Is there somewhere that the team has a roadmap? Thanks very much for sharing your hard work!

samouri:
Hey @Slapbox, 

We don't have any features on the roadmap for now.

I believe we already have `Node.remove` implemented:

https://github.com/ampproject/worker-dom/blob/5aa157976f692c760eae65cdb375523f86d1e129/src/worker-thread/dom/Node.ts#L375-L383

We likely just need to update the compat table.

Regarding `replaceWith`, I don't think we have that implemented. We are always open to outside contribution in case you'd like to create a pull request adding support!

Slapbox:
Thanks very much for your reply @samouri! I'd love to contribute to getting those features included, but I wouldn't even know where to begin.

samouri:
I believe the new method `replaceWith` would go in `Node.ts` file. You should follow the spec as [explained in mdn](https://developer.mozilla.org/en-US/docs/Web/API/ChildNode/replaceWith). Particularly useful is the Polyfill.

You'd then want to create a new file called `src/test/node/replaceWith.test.ts` and follow a similar pattern to other methods, like https://github.com/ampproject/worker-dom/blob/main/src/test/node/remove.test.ts

kristoferbaxter:
I haven't made a contribution for a while to this repo, so I'll take this on.

samouri:
I'll perform a release on Monday

Slapbox:
Holy moly, thanks @kristoferbaxter for your PR and @samouri for your guidance and maintenance!

kristoferbaxter:
Released in `0.28.0`.

PR #1028, opened 2021-03-12
Add Node.replaceWith capability

kristoferbaxter:
Implements functionality requested in #1023.


src/test/node/replaceWith.test.ts
```diff
@@ -0,0 +1,125 @@
...
+  t.context = {
+    document,
+    parent: document.createElement("div"),
+    child: document.createElement("div"),
+    childTwo: document.createElement("div"),
+    childThree: document.createElement("div"),
+  };
+});
+
+test('replacing unattached child results in no changes', (t) => {
```
samouri, line 42: One test missing: `replacing child with nothing`. Technically a 0-args call to `replaceWith` should be equivalent to node.remove()

samouri (changes requested):
Great work. One nit about potentially missing edge case

src/test/node/replaceWith.test.ts
kristoferbaxter: Added test. Good callout!

samouri (approved):

Issue #1009, opened 2020-11-23
click() not working on HTMLAnchorElement


steve-mills:
Hi guys,
The HTMLElement.click() doesn't seem to be working if I get an element from the existing DOM with getElementById.
Do I need to dynamically add an element to the Document get this to work? 

Here's an example of what I have tried. The setAttribute method works fine when used like this, .click() doesn't.

document.querySelectorAll('.button').addEventListener('click', clicker);
const clicker = function(event){ 
 let ev = event.target.id;
let anchor = ev + '.html';
 let evB = ev + 'Button';
 let evBL = document.getElementById(evB);
 evBL.click();
   };

Thanks for your help,
Kind Regards

samouri:
@steve-mills: do you know if the `evBL` click handler is called or not? Also, what type of element is `evBL`?
Calling `.click()` on anchor tags (`<a>`) is currently broken.

steve-mills:
Hi Jake,
Thanks heaps for your response. 
Sorry evBL is an anchor tag (I'm not a developer and I'm new to git). 
event.target.id is a hex shaped SVG path that triggers an anchor when clicked. I should have commented
. Is there a work around to trigger a href?

samouri:
Ok! Glad to hear it is indeed an anchor.
I don't think fixing it is at the top of anyone's priority list right now.
We are of course open to pull requests!

samouri:
Closing this as a dupe of https://github.com/ampproject/worker-dom/issues/917

PR #1033, opened 2021-03-22
Added partial support for Element.scrollIntoView()

andrei0x309:
Since there isn't a way now to programmatically scroll to an element, I thought it would be nice if there was some basic support for that.

I've seen some issues like [issue-1009](https://github.com/ampproject/worker-dom/issues/1009)  that can be resolved if `worker-dom` supports `scrollIntoView()`, I only added basic support(no support for method params) since it required less code, also the 'smooth' option today is supported by using CSS in most browsers, and those that don't support CSS smooth, they don't support `scrollIntoView()` with smooth as an option.

I couldn't write a test since I believe `Window.pageY` was necessary and is not supported by `worker-dom`, but I wrote a demo instead of a test. 

Thanks, and I appreciate any feedback.


samouri:
@andrei0x309, thank you so much for the PR! At initial glance this looks great. I'll give it a more thorough review later today.

One quick note: we usually write unit tests that ensure messages passed between main-thread and worker-thread are as expected, as opposed to making them integration tests (which could involve checking `Window.pageY`).

Therefore potential tests you could write:
1. Ensure worker-thread sends the expected message to the main-thread when Element.scrollIntoView is called.
2. Ensure main-thread calls the expected DOM API when receiving the correct message from the worker-thread


Here's an example of a test expecting mutation events to be sent from the worker-thread:
https://github.com/ampproject/worker-dom/blob/f5da1e6231d552618290ae07cdfd4ac547b3ff12/src/test/mutation-transfer/amp.test.ts#L42-L57

andrei0x309:
@samouri Thanks, for the reply, 

I added the transfer mutation test for scrollIntoView().

src/transfer/TransferrableMutation.ts
```diff
@@ -44,6 +45,7 @@ export const isUserVisibleMutation = (type: TransferrableMutationType): boolean
     case TransferrableMutationType.STORAGE:
     case TransferrableMutationType.OFFSCREEN_CANVAS_INSTANCE:
     case TransferrableMutationType.FUNCTION_CALL:
+    case TransferrableMutationType.SCROLL_INTO_VIEW:
```
samouri, line 48: I know the comment describes user-visible changes to _DOM_, but I actually think the intention is any user-visible change (which would include this). Can you delete this line s.t. it returns true
```suggestion
```

src/test/mutation-transfer/scrollIntoView.test.ts
```diff
@@ -0,0 +1,51 @@
...
+ * limitations under the License.
+ */
+
+ import anyTest, { TestInterface } from 'ava';
+ import { Document } from '../../worker-thread/dom/Document';
+ import { TransferrableMutationType } from '../../transfer/TransferrableMutation';
+ import { createTestingDocument } from '../DocumentCreation';
+ import { expectMutations } from '../Emitter';
+ import { Element } from '../../worker-thread/dom/Element';
+import { TransferrableKeys } from '../../transfer/TransferrableKeys';
```
samouri, line 23: nit: did all of the prepush checks run? we normally run `prettier` over the codebase, and the whitespace here doesn't look right.

demo/scroll-into-view/scroll-into-view.js
```diff
@@ -0,0 +1,7 @@
+document.getElementById('scroll-button').addEventListener('click', () => {
+  document.getElementById('scroll-element').scrollIntoView();
+  });
```
samouri, line 3: nit: whitespace here doesn't look right. 
```suggestion
});
```

demo/scroll-into-view/scroll-into-view.js
```diff
@@ -0,0 +1,7 @@
...
+
+document.getElementById('scroll-button-up').addEventListener('click', () => {
+  document.getElementById('scroll-button').scrollIntoView();
+  });
```
samouri, line 7:
```suggestion
});  
```

demo/scroll-into-view/index.html
```diff
@@ -0,0 +1,48 @@
...
+    <link href="/demo.css" rel="stylesheet" />
+    <script src="/dist/main.mjs" type="module"></script>
+    <script src="/dist/main.js" nomodule defer></script>
+    <!-- This comment block is intended to make it easier to test both the script module and nomodule path -->
+    <!-- Comment either block to enable module/nomodule or disable it. -->
+    <!-- <script src="/dist/main.js" defer></script> -->
+  </head>
+  <body>
+    <div src="/scroll-into-view/scroll-into-view.js" id="upgrade-me">
+      <p>Demo of <b>element.scrrollIntoView()</b></p>
```
samouri, line 16:
```suggestion
      <p>Demo of <b>element.scrollIntoView()</b></p>
```

samouri (changes requested):
Excellent PR :). Only one blocking request (to ensure it is considered user visible).
Otherwise LGTM.

As an action item on my side I'm going to make a new command, `yarn format` for automatically running prettier on files

src/transfer/TransferrableMutation.ts
andrei0x309: Yeah, you're completely right I only thought of DOM and not considered the scroll changing position as a user-visible change which it actually is.

src/test/mutation-transfer/scrollIntoView.test.ts
andrei0x309: This time I didn't. 
Sorry for that.

samouri (approved):

PR #1035, opened 2021-03-24
storage fixes

samouri:
**summary**
Fixes two bugs discovered in https://github.com/ampproject/amphtml/issues/32984. Note that both fixes are only **5 LOC**. The rest are tests. 

1. When `AMP.getState()` is called with no args, the key was recorded as undefined in worker-thread but empty string in main-thread. The fix was to coerce falsy keys to an empty string in worker-thread as well. Without this the messages would never be matched properly and worker-thread always resolved the request with `null`.
2. When the `storageKey` being requested is the 0th indexed string in StringContext, the StorageProcessor would always send over null. Essentially never a problem for a dom post-hydration, but could easily happen via `nodom`.

As a bonus change, I've also added 'GET_STORAGE' to the list of debuggable events.

### DIFF
--- a/src/main-thread/commands/storage.ts
+++ b/src/main-thread/commands/storage.ts
@@ -23,12 +23,12 @@ import { MessageType, StorageValueToWorker, GetOrSet } from '../../transfer/Mess
 export const StorageProcessor: CommandExecutorInterface = (strings, nodeContext, workerContext, objectContext, config) => {
   const allowedExecution = config.executorsAllowed.includes(TransferrableMutationType.STORAGE);
 
-  const get = (location: StorageLocation, key: string | null): void => {
+  const get = (location: StorageLocation, key: string): void => {
     if (config.sanitizer && location === StorageLocation.AmpState) {
       config.sanitizer.getStorage(location, key).then((value) => {
         const message: StorageValueToWorker = {
           [TransferrableKeys.type]: MessageType.GET_STORAGE,
-          [TransferrableKeys.storageKey]: key || '',
+          [TransferrableKeys.storageKey]: key,
           [TransferrableKeys.storageLocation]: location,
           [TransferrableKeys.value]: value,
         };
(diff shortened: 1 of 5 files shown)


src/test/main-thread/commands/storage.test.ts
```diff
@@ -0,0 +1,83 @@
...
+test('StorageProcessor sends storage value event to worker', async (t) => {
+  const { processor, messages } = getStorageProcessor(['key']);
+  const mutation: number[] = [];
+  mutation[StorageMutationIndex.Operation] = GetOrSet.GET;
+  mutation[StorageMutationIndex.Location] = StorageLocation.AmpState;
+  mutation[StorageMutationIndex.Key] = 0;
+  const mutations = new Uint16Array(mutation);
+
+  processor.execute(mutations, 0, true);
+  await Promise.resolve(setTimeout);
```
samouri, line 73: Note: ideally this test could receive `getNextMessage` as an async iterator rather than this hacky macroTask wait.

I looked into it and would require fairly complex code. Will continue researching as a followup since a very similar need exists in multiple test.

src/main-thread/commands/storage.ts
```diff
@@ -98,8 +98,8 @@ export const StorageProcessor: CommandExecutorInterface = (strings, nodeContext,
       const keyIndex = mutations[startPosition + StorageMutationIndex.Key];
       const valueIndex = mutations[startPosition + StorageMutationIndex.Value];
 
-      const key = keyIndex > 0 ? strings.get(keyIndex) : null;
-      const value = valueIndex > 0 ? strings.get(valueIndex) : null;
+      const key = keyIndex >= 0 ? strings.get(keyIndex) : null;
+      const value = valueIndex >= 0 ? strings.get(valueIndex) : null;
```
kristoferbaxter, line 102: This is one of the two hardest things in computer science after all.

src/test/main-thread/commands/storage.test.ts
kristoferbaxter: A followup would be great, and potentially a good task for someone new to the project.

src/worker-thread/amp/amp.ts
```diff
@@ -35,6 +35,7 @@ export class AMP {
    * @param key
    */
   getState(key: string): Promise<{} | null> {
+    key = key || '';
```
kristoferbaxter, line 38: Just for clarity, do you want this to be a default value here or in the function arguments?

`getState(key: string = ''): Promise<{} | null> {`

kristoferbaxter (approved):

src/worker-thread/amp/amp.ts
samouri: Good idea. I'll update

PR #1042, opened 2021-03-29
Introduce sandboxed iframe mode.

samouri:
**background**
The primary use-case of `worker-dom` is to power [`<amp-script>`](https://amp.dev/documentation/components/amp-script/).  With the current mechanisms in place, it acts well for 1p (mostly trusted) scripts. `amp-script` now has a new [use case](https://github.com/ampproject/amphtml/issues/30193) to allow less-trusted 3p scripts to run on the page without a CSP, and for this we need a stronger security boundary.

The goal of this PR is to enable a "sandboxed" mode by creating the Worker within a sandboxed iframe instead of within the main frame.

**design**

When starting up, `worker-dom` will create an iframe instead of a WebWorker. This iframe needs to perform a handshaking process and then essentially act as a proxy between the subframe Worker and the main frame. To make implementation as simple as possible, I've created a class called `IframeWorker` that presents almost the exact same API surface as a standard Worker.

The iframe must follow a specific contract:
1. When initialized, send out a message `{type: 'iframe-ready'}`
2. When receiving a message with shape: `{type: 'init-worker', code: string}`, create a Worker with the specified code.
3. Proxy all messages received via shape `{type: 'postMessage', message: string}` to the subframe Worker.
4.  Proxy all messages received from the Worker back to the parent. The shape of should be: `{type: 'onmessage' | 'onerror', message: string }`

In order to take advantage of this feature, consumers will need to use the new option in the `upgradeElement` API called `sandboxed`. The option may remain undefined to men _off_, or be an object with the key `iframeUrl`. The specified iframe must follow the contract described above. An example is provided in the demo folder.

**hosting the iframe**
Depending on whether we want to release this for non-AMP mode as well, we could also have a default value for `iframeUrl` that points to a Github Pages hosted iframe.
 
 ~~**note**: There is some drift in our src code alignment in `prettier` and I've extracted that to a small separate PR: https://github.com/ampproject/worker-dom/pull/1041~~

### DIFF
--- a/demo/sandboxed/sandbox-iframe.html
+++ b/demo/sandboxed/sandbox-iframe.html
@@ -1,0 +1,65 @@
+<html>
+  <script>
+    // The IframeWorker contract is that it must:
+    // 1. Send a ready message.
+    // 2. Listen for the Worker code to initialize it with.
+    // 3. Proxy all messages between the Worker and Parent, including errors.
+
+    let parentOrigin = '*';
+    const MESSAGE_TYPES = {
+      ready: 'iframe-ready',
+      init: 'init-worker',
+      onmessage: 'onmessage',
+      onerror: 'onerror',
+      onmessageerror: 'onmessageerror',
+      postMessage: 'postMessage',
+    };
+    function send(type, message) {
+      if (type !== MESSAGE_TYPES.ready && parentOrigin === '*') {
+        throw new Error('Broadcast banned except for iframe-ready message.');
+      }
+      parent.postMessage({ type, message }, parentOrigin);
+    }
+
+    function listen(type, handler) {
+      window.addEventListener('message', (event) => {
+        if (event.source !== parent) {
+          return;
+        }
+        parentOrigin = event.origin;
+
+        if (event.data.type === type) {
+          handler(event.data);
+        }
+      });
+    }
+
+    // Send initialization.
+    send(MESSAGE_TYPES.ready);
+
+    let worker = null;
+    // Listen for Worker Init.
+    listen(MESSAGE_TYPES.init, ({ code }) => {
+      if (worker) {
+        return;
+      }
+      worker = new Worker(URL.createObjectURL(new Blob([code])));
+
+      // Proxy messages Worker to parent Window.
+      worker.onmessage = (e) => send(MESSAGE_TYPES.onmessage, e.data);
+      worker.onmessageerror = (e) => send(MESSAGE_TYPES.onmessageerror, e.data);
+      worker.onerror = (e) =>
+        send(MESSAGE_TYPES.onerror, {
+          lineno: e.lineno,
+          colno: e.colno,
+          message: e.message,
+          filename: e.filename,
+        });
+
+      // Proxy message from parent Window to Worker.
+      listen(MESSAGE_TYPES.postMessage, ({ message }) =>
+        worker.postMessage(message)
+      );
+    });
+  </script>
+</html>
--- a/demo/sandboxed/sandbox-iframe.html
+++ b/demo/sandboxed/sandbox-iframe.html
@@ -1,10 +1,8 @@
 <html>
   <script>
-    // The IframeWorker contract is that it must:
-    // 1. Send a ready message.
-    // 2. Listen for the Worker code to initialize it with.
-    // 3. Proxy all messages between the Worker and Parent, including errors.
-
+    /**
+     * See `iframe-worker.ts` for the iframe proxy contract.
+     */
     let parentOrigin = '*';
     const MESSAGE_TYPES = {
       ready: 'iframe-ready',
--- a/src/main-thread/iframe-worker.ts
+++ b/src/main-thread/iframe-worker.ts
@@ -24,6 +24,12 @@ export type MessageToIframe = { type: 'terminate' } | { type: 'init-worker'; cod
 /**
  * An almost drop-in replacement for a standard Web Worker, although this one
  * within a sandboxed cross-origin iframe for a heightened security boundary.
+ * For more details on Worker, see: https://developer.mozilla.org/en-US/docs/Web/API/Worker
+ *
+ * The iframe used for sandboxing must follow a specific contract. It:
+ *   1. Must send a ready message to the main-thread.
+ *   2. Must listen for a message from main-thread with the code to initialize a Worker with.
+ *   3. Must proxy all messages between the Worker and Parent, including errors.
  */
 class IframeWorker {
   // Public Worker API
@@ -33,9 +39,12 @@ class IframeWorker {
 
   // Internal variables.
   private iframe: HTMLIFrameElement;
-  private url: string | URL;
 
-  constructor(url: string | URL, iframeUrl: string) {
+  /**
+   * @param url The URL to initiate the worker from.
+   * @param iframeUrl The URL of the iframe to use as the worker proxy.
+   */
+  constructor(private url: string | URL, iframeUrl: string) {
     this.iframe = window.document.createElement('iframe');
     this.iframe.setAttribute('sandbox', 'allow-scripts');
     this.iframe.setAttribute('style', 'display:none');
@@ -80,16 +89,24 @@ class IframeWorker {
     });
   }
 
+  /**
+   * See https://developer.mozilla.org/en-US/docs/Web/API/Worker/postMessage
+   * @param message
+   * @param transferables
+   */
   postMessage(message: any, transferables?: Array<Transferable>) {
     const msg: MessageToIframe = { type: 'postMessage', message };
     this.iframe.contentWindow!.postMessage(msg, '*', transferables);
   }
 
+  /**
+   * See https://developer.mozilla.org/en-US/docs/Web/API/Worker/terminate
+   */
   terminate() {
     const msg: MessageToIframe = { type: 'terminate' };
     this.iframe.contentWindow!.postMessage(msg, '*');
     this.iframe.remove();
-  } 
+  }
 }
 
 export { IframeWorker };
--- a/src/main-thread/iframe-worker.ts
+++ b/src/main-thread/iframe-worker.ts
@@ -1,0 +1,105 @@
+/**
+ * Copyright 2021 The AMP HTML Authors. All Rights Reserved.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS-IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+type MessageFromWorker = {
+  type: 'onmessage' | 'onerror' | 'onmessageerror';
+  message: any;
+};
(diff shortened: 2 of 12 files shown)


samouri:
As we only know of a use case for this for `amp-script`, it doesn't make sense to bloat the regular binaries with the code for IframeWorker as well.

Is there a way to conditionally import `IframeWorker` within `worker.ts` in only AMP modes?

kristoferbaxter:
You can create a new constant for AMP only additions, and use replacement to filter them out during the compilation process.

Or, you can create a custom plugin for removing the logic (see how debug is removed from evaluators).

samouri:
@kristoferbaxter: was just about to push when you commented that. I was worried that a const wouldn't be enough to DCE the entire import, but it looks like it worked!

**before**
```
./dist/main.js 5.33 kb  
  ✔ ./dist/main.mjs 4.36 kb
```

**after**
ours:
```
./main.js 4.67 kb  
./dist/main.mjs 3.83 kb 
```

src/test/main-thread/iframe-worker.test.ts
```diff
@@ -0,0 +1,122 @@
...
+}>;
+
+test.beforeEach((t) => {
+  const env = new Env();
+  const sentToIframe: MessageToIframe[] = [];
+
+  const messageListeners: any = [];
+  env.window.addEventListener = (type: string, listener: any) => messageListeners.push(listener);
+
+  global.fetch = (blob: any) =>
```
kristoferbaxter, line 35: Believe globals are cleaned up automatically, but can we verify?

kristoferbaxter (approved):
Looks good to me, thanks for ensuring this is specific to the AMP output too.

Added some comments and nits to consider.

src/main-thread/iframe-worker.ts
```diff
@@ -0,0 +1,105 @@
...
+export type MessageFromIframe = { type: 'iframe-ready' } | MessageFromWorker;
+export type MessageToIframe = { type: 'terminate' } | { type: 'init-worker'; code: string } | { type: 'postMessage'; message: any };
+
+/**
+ * An almost drop-in replacement for a standard Web Worker, although this one
+ * within a sandboxed cross-origin iframe for a heightened security boundary.
+ */
+class IframeWorker {
+  // Public Worker API
+  public onerror: (this: IframeWorker, ev: ErrorEvent) => any;
```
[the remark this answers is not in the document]
samouri: I copy/pasted from the tsc definition of `Worker`

src/main-thread/iframe-worker.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  // Internal variables.
+  private iframe: HTMLIFrameElement;
+  private url: string | URL;
+
+  constructor(url: string | URL, iframeUrl: string) {
+    this.iframe = window.document.createElement('iframe');
+    this.iframe.setAttribute('sandbox', 'allow-scripts');
+    this.iframe.setAttribute('style', 'display:none');
+    this.iframe.setAttribute('src', iframeUrl);
```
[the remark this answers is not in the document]
samouri: I didn't know there was a difference!

src/main-thread/iframe-worker.ts
```diff
@@ -0,0 +1,105 @@
...
+    this.url = url;
+
+    this.setupInit();
+    this.proxyFromWorker();
+    window.document.body.appendChild(this.iframe);
```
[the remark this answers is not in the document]
samouri: I don't need it at runtime but I do need it for the tests to grab the overriden Document

src/main-thread/iframe-worker.ts
```diff
@@ -0,0 +1,105 @@
...
+
+  postMessage(message: any, transferables?: Array<Transferable>) {
+    const msg: MessageToIframe = { type: 'postMessage', message };
+    this.iframe.contentWindow!.postMessage(msg, '*', transferables);
+  }
+
+  terminate() {
+    const msg: MessageToIframe = { type: 'terminate' };
+    this.iframe.contentWindow!.postMessage(msg, '*');
+    this.iframe.remove();
```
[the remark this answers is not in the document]
samouri: Correct! A Worker can never be unterminated

src/main-thread/iframe-worker.ts
```diff
@@ -0,0 +1,105 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+type MessageFromWorker = {
+  type: 'onmessage' | 'onerror' | 'onmessageerror';
+  message: any;
+};
+export type MessageFromIframe = { type: 'iframe-ready' } | MessageFromWorker;
+export type MessageToIframe = { type: 'terminate' } | { type: 'init-worker'; code: string } | { type: 'postMessage'; message: any };
```
rcebulko, line 22: nit: Not sure if linting/prettier is set up in this repo, but breaking this union across lines would be nice

src/main-thread/iframe-worker.ts
```diff
@@ -0,0 +1,105 @@
...
+  // Public Worker API
+  public onerror: (this: IframeWorker, ev: ErrorEvent) => any;
+  public onmessage: (this: IframeWorker, ev: MessageEvent) => any;
+  public onmessageerror: (this: IframeWorker, ev: MessageEvent) => any;
+
+  // Internal variables.
+  private iframe: HTMLIFrameElement;
+  private url: string | URL;
+
+  constructor(url: string | URL, iframeUrl: string) {
```
rcebulko, line 38:
```suggestion
  constructor(private url: string | URL, private iframeUrl: string) {
```

src/main-thread/iframe-worker.ts
```diff
@@ -0,0 +1,105 @@
...
+    window.addEventListener('message', (event: MessageEvent) => {
+      if (event.source != this.iframe.contentWindow) {
+        return;
+      }
+
+      const { type, message } = event.data as MessageFromWorker;
+      if (type == 'onmessage' && this.onmessage) {
+        this.onmessage({ ...event, data: message });
+      } else if (type === 'onerror' && this.onerror) {
+        this.onerror({ ...message });
```
rcebulko, line 76: What can't this just be `this.onerror(message)`?

src/main-thread/worker.ts
```diff
@@ -80,7 +83,11 @@ export class WorkerContext {
       }).call(self);
       ${authorScript}
       //# sourceURL=${encodeURI(config.authorURL)}`;
-    this[TransferrableKeys.worker] = new Worker(URL.createObjectURL(new Blob([code])));
+    if (!config.sandbox) {
+      this[TransferrableKeys.worker] = new Worker(URL.createObjectURL(new Blob([code])));
+    } else if (IS_AMP && config.sandbox){
```
rcebulko, line 88: By this point in the conditional, `!config.sandbox` was false so `config.sandbox` should be truthy. 
```suggestion
    } else if (IS_AMP) {
```

src/main-thread/worker.ts
```diff
@@ -80,7 +83,11 @@ export class WorkerContext {
...
+      this[TransferrableKeys.worker] = new IframeWorker(URL.createObjectURL(new Blob([code])), config.sandbox.iframeUrl) as Worker;
```
rcebulko, line 89: Should `IframeWorker` extend `Worker` so explicit casts aren't necessary? Or is that not...Work-able? 🥁

src/main-thread/iframe-worker.ts
samouri: It is setup, but set to 150 width. Feel free to open a PR that changes it to lower :)

src/main-thread/iframe-worker.ts
```diff
@@ -0,0 +1,105 @@
...
+  }
+
+  terminate() {
+    const msg: MessageToIframe = { type: 'terminate' };
+    this.iframe.contentWindow!.postMessage(msg, '*');
+    this.iframe.remove();
+  }
+
+  addEventListener() {
+    throw new Error('Unsupported operation');
```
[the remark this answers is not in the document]
samouri: Actually, I forgot that we export `ExportedWorker` instead of the standard `Worker`. We can remove these :)

demo/sandboxed/sandbox-iframe.html
```diff
@@ -0,0 +1,65 @@
+<html>
+  <script>
+    // The IframeWorker contract is that it must:
+    // 1. Send a ready message.
+    // 2. Listen for the Worker code to initialize it with.
+    // 3. Proxy all messages between the Worker and Parent, including errors.
```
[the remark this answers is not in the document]
samouri: Good thinking! I agree. I'll copy it there

src/main-thread/iframe-worker.ts
rcebulko: I'm pretty sure? It's distinctly possible I'm mixing up variants of TypeScript, ES6, C++, idk Python etc but I feel like TS is the one that supports this.

src/main-thread/iframe-worker.ts
samouri: On I think it can be. I changed that line many times and ended up with one that isn't sensical :)

src/main-thread/worker.ts
samouri: Thats true! I think it is less obvious to a reader though which is why I intentionally left this in. Don't feel too strongly, I'll update

src/main-thread/worker.ts
samouri: I have absolutely no idea what happens if you extend a builtin like `Worker`. It definitely isn't a great idea in this case where the implementation is so drastically different. Would make more sense to extend the interface if that were possible.

Have a different solution that I think you'd approve of (worker: Worker | IframeWorker)

src/test/main-thread/iframe-worker.test.ts
```diff
@@ -0,0 +1,122 @@
...
+    Promise.resolve({
+      text: () => Promise.resolve(blob.toString().slice('BLOB:'.length)),
+    }) as any;
```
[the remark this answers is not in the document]
samouri: I feel weird about this 🤔

src/test/main-thread/iframe-worker.test.ts
```diff
@@ -0,0 +1,122 @@
...
+  const worker = new IframeWorker(URL.createObjectURL(new Blob(['worker code'])), 'https://example.com' as any);
```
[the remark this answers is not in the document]
samouri: I don't anymore but had in the past when I had `IframeWorker` exported in the `Worker` interface. Fixing

src/main-thread/worker.ts
rcebulko: There's a way to define an interface/typedef and declare that both Worker and IframeWorker implement it, but idk how to do that and doubt it's worth the time atm

src/main-thread/iframe-worker.ts
samouri: Oh. It works for `url`, but not for the iframe (which is not an arg)

samouri: Updated for `url`

src/test/main-thread/iframe-worker.test.ts
samouri: In ava 3.x we are safe. may need to change things when we upgrade to 4.x.
I'll do the migration asap, with keeping the process isolation intact

https://github.com/avajs/ava/blob/main/docs/01-writing-tests.md#test-isolation

src/main-thread/iframe-worker.ts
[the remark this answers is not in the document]
samouri: Done

PR #1047, opened 2021-04-20
WorkerDOM.upgrade should wait for Worker init.

samouri:
**summary**
Regular web workers are initialized synchronously, whereas the new `IframeWorker` introduced recently is initialized async. In order to ensure all sent messages end up received we have two options:
1. Buffer all received messages within `IframeWorker` and send once it has been completely initialized.
2. Force the `WorkerDOM.upgrade` to only resolve once the worker is ready.

I've opted to do both in this PR 🤷


src/main-thread/worker.ts
```diff
@@ -94,6 +94,14 @@ export class WorkerContext {
     }
   }
 
+  /**
+   * Returns a Promise that resolves when the Worker is ready to receive messages.
+   * @returns {Promise<void>}
+   */
+  ready() {
+    return (this.worker as IframeWorker).readyPromise || Promise.resolve();
```
rcebulko, line 102: In what circumstance would `readyPromise` be falsey if it's being initialized in the constructor?

src/main-thread/iframe-worker.ts
```diff
@@ -65,11 +71,14 @@ class IframeWorker {
       fetch(this.url.toString())
         .then((res) => res.text())
         .then((code) => {
-          if ((event.data as MessageFromIframe).type == 'iframe-ready') {
+          const data = event.data as MessageFromIframe;
+          if (data.type == 'iframe-ready') {
```
rcebulko, line 75: Is there a reason to use a string literal instead of `MESSAGE_TYPES.ready` etc here? Wondering specifically if the latter might get compiled/minified better

rcebulko (approved):
Left a couple Qs but generally looks good to me

src/main-thread/iframe-worker.ts
samouri: I don't think I have the struct for it here. There's less of a need in TS when it is properly typed, whereas in JS its helpful to avoid typos

src/main-thread/worker.ts
samouri: `Worker` will never have `readyPromise` (as well as WorkerImpl which is used in tests), whereas `IframeWorker` will.

src/main-thread/iframe-worker.ts
rcebulko: Is there typechecking on equality comparisons like this? I don't think there would be?

samouri: Yep! The error looks like this:
```
This condition will always return 'false' since the types ... have no overlap
```

PR #1066, opened 2021-06-15
Consolidate prettier

samouri:
**summary**
- Commit e1077b236160a87c7b9d8be4d3b21582ccf339a7 deletes `config/.prettierrc`, `tools/linter/.prettierrc`, and `tools/linter/config/.prettierrc`. It adds a root-level config, and also removes the printWidth option. I personally prefer root-level because then `prettier` can find it without a --config flag, as well as VSCode.
- Commit fce9291bd6786a968a89f214d7c05d9702b94aa6 then runs prettier over the whole `src` and `tools` dirs.


kristoferbaxter:
No solid thoughts on consolidating the prettier configurations, but can we keep the line length the same as it was?

samouri:
> No solid thoughts on consolidating the prettier configurations, but can we keep the line length the same as it was?

This PR was largely a result of a complaint around long lines 🙃

samouri:
I personally care much more about consolidating the files and moving to toplevel to let my IDE pick it up and less so about the specific prettier options.

I've elected to pick a configuration that has the fewest changes (same line length, bracket spacing, etc.)

kristoferbaxter (approved):

PR #1067, opened 2021-06-18
Separate out all type-only imports

samouri:
**summary**
Separates out type-only imports from regular imports. 
Helps ensure side effects don't get needlessly added.

```
- dist/worker/worker.mjs: 11.88 kb --> 11.79 kb
- dist/worker/worker.js: 13.51 kb --> 13.42 kb
- dist/amp-production/worker/worker.mjs: 12.81 kb --> 12.68 kb
- dist/amp-production/worker/worker.js:  14.39 kb --> 14.29 kb
```

cc @kristoferbaxter


rcebulko (approved):
This looks like a solid step forward! I'm curious, how many of these type imports are importing classes as type signatures versus typedefs, aliases, interfaces, etc that don't have a specific class implementation behind them?

PR #1072, opened 2021-07-02
worker: fix faulty import type

samouri:
**summary**
Recent PR https://github.com/ampproject/worker-dom/pull/1067 had an error. The worker-thread `index.ts` and `index.amp.ts` should not use `import type` for imports whose values are actually used!

I'm incredibly surprised that `TypeScript` didn't catch this, and opened up an issue: https://github.com/microsoft/TypeScript/issues/44868

**testing done**

Manually ran `yarn demo` and checked each example


kristoferbaxter (approved):

samouri:
This wasn't ready yet! I had planned on reverting all of the type imports. Followup in https://github.com/ampproject/worker-dom/pull/1073

PR #1074, opened 2021-07-12
Babel plugins: remove outdated proposals

samouri:
**summary**

We recently started receiving a new warning during `worker-dom` build:

```
Though the "loose" option was set to "false" in your @babel/preset-env config, it will not be used for @babel/plugin-proposal-private-methods since the "loose" mode option was set to "true" for @babel/plugin-proposal-class-properties.
The "loose" option must be the same for @babel/plugin-proposal-class-properties, @babel/plugin-proposal-private-methods and @babel/plugin-proposal-private-property-in-object (when they are enabled): you can silence this warning by explicitly adding
        ["@babel/plugin-proposal-private-methods", { "loose": true }]
to the "plugins" section of your Babel config.
```

This PR fixes that warnings by removing our explicit inclusion of two proposals which are automatically configured by preset-env anyway.

---

I also made an unrelated changes:  migrate from `rollup-plugin-babel` to `@rollup/plugin-babel` (new location for the same package)

**testing done**

I checked via `diff` and the output binaries are byte for byte identical.

cc @jridgewell since this is a change to babel config.


config/rollup.plugins.js
```diff
@@ -28,10 +28,11 @@ const walk = require('acorn-walk');
  * - allowPostMessage Should we allow postMessage to/from the Worker?
  */
 export function babelPlugin({ transpileToES5, allowConsole = false, allowPostMessage = true }) {
-  const targets = transpileToES5 ? { browsers: ['last 2 versions', 'ie >= 11', 'safari >= 7'] } : { esmodules: true };
+  const targets = transpileToES5 ? { browsers: ['ie >= 11', 'safari >= 8'] } : { esmodules: true };
```
jridgewell, line 31: I believe this means we're only targeting Safari 9 and Edge 12 in our matrix, which means, eg, a Chrome bug won't be transpiled down appropriately.

samouri: WDYT of `"> 1.0%"`

jridgewell (approved):

PR #1076, opened 2021-07-13
babel: update browserlist

samouri:
**summary**
Modifying the browserlist according to recommendation by [jamie.build](https://jamie.build/last-2-versions)


kristoferbaxter:
The ES5 build is designed to work on IE11, so having IE11 in the list is intentional.

Is there a reason for making a change here versus recommending people use the mjs build if they want newer syntax output?

PR #1091, opened 2021-09-08
nodom: workerDOM should be document.defaultView

samouri:
**summary**
Fixes https://github.com/ampproject/amphtml/issues/35902

1. Renames `DocumentLite.ts` to `DocumentStub.ts`. The filename should match the exported class :)
2. Modifies `workerDOM` export to actually be the `document.defaultView` which is essential for propagating properties (localStorage/sessionStorate) to the global `self`. This is also how the regular binary works. See below snippets

https://github.com/ampproject/worker-dom/blob/11ff429516e4dcfb4f64794bed5f52163d5d9166/src/main-thread/worker.ts#L80

https://github.com/ampproject/worker-dom/blob/11ff429516e4dcfb4f64794bed5f52163d5d9166/src/worker-thread/index.amp.ts#L137

**testing done**
Copied the built binaries over to a local `amphtml` and confirmed that localStorage access was available on `self`

### DIFF
--- a/src/worker-thread/MutationTransfer.ts
+++ b/src/worker-thread/MutationTransfer.ts
@@ -22,7 +22,7 @@ import { Node } from './dom/Node';
 import { Phase } from '../transfer/Phase';
 import { phase, set as setPhase } from './phase';
 import { Document } from './dom/Document';
-import { DocumentStub } from './dom/DocumentLite';
+import { DocumentStub } from './dom/DocumentStub';
 
 let pending = false;
 let pendingMutations: Array<number> = [];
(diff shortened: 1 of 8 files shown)


rcebulko:
> https://github.com/ampproject/worker-dom/blob/11ff429516e4dcfb4f64794bed5f52163d5d9166/src/main-thread/worker.ts#L80

Why can't this be `Object.assign(self, workerDOM);`?

samouri:
> > https://github.com/ampproject/worker-dom/blob/11ff429516e4dcfb4f64794bed5f52163d5d9166/src/main-thread/worker.ts#L80
> 
> Why can't this be `Object.assign(self, workerDOM);`?

Probably can be! Feel free to make a separate PR!

rcebulko (approved):

PR #1100, opened 2021-09-28
server-lib bugfix: need scopingRoot

samouri:
Followup fix to a lil bug in https://github.com/ampproject/worker-dom/pull/1096.
Note: the bug only affects server-lib.

---

Followup: need to come up with a good system for parameterizing tests

### DIFF
--- a/src/test/htmlelement/querySelectorAll.test.ts
+++ b/src/test/htmlelement/querySelectorAll.test.ts
@@ -45,11 +45,10 @@ test.serial('server=true: test Element.querySelectorAll on id selectors', (t) =>
   try {
     const document = createTestingDocument();
     const testEl = document.createElement('test-el');
-    const id = 'TEST_EL_ID';
-    testEl.setAttribute('id', id);
+    testEl.setAttribute('id', 'TEST_EL_ID');
     document.appendChild(testEl);
 
-    const result = document.querySelectorAll(`#${id}`);
+    const result = document.querySelectorAll('#TEST_EL_ID');
     t.deepEqual(result, [testEl]);
   } finally {
     delete process.env.SERVER;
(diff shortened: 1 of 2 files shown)


src/test/htmlelement/querySelectorAll.test.ts
```diff
@@ -39,6 +39,23 @@ test('test Element.querySelectorAll on id selectors', (t) => {
   t.deepEqual(document.querySelectorAll(`#${DIV_ID}`), [div]);
 });
 
+// TODO: create better solution for parameterizing server tests.
+test.serial('server=true: test Element.querySelectorAll on id selectors', (t) => {
+  process.env.SERVER = 'true';
```
rcebulko, line 44: not sure what testing framework is in use, but maybe something like `sandbox.stub(process.env, 'SERVER').value('true');` or something similar could do the trick.

src/test/htmlelement/querySelectorAll.test.ts
```diff
@@ -39,6 +39,23 @@ test('test Element.querySelectorAll on id selectors', (t) => {
...
+  try {
+    const document = createTestingDocument();
+    const testEl = document.createElement('test-el');
+    const id = 'TEST_EL_ID';
+    testEl.setAttribute('id', id);
+    document.appendChild(testEl);
+
+    const result = document.querySelectorAll(`#${id}`);
```
rcebulko, line 52: nit: I'm usually a fan of DRY "do repeat yourself" for tests, in this case replicating the string in multiple places

rcebulko (approved):

src/test/htmlelement/querySelectorAll.test.ts
samouri: I'd call that WET! I'll update

src/test/htmlelement/querySelectorAll.test.ts
samouri: Will investigate this for a future PR

PR #1101, opened 2021-10-01
Added a flag to selectively use shadow dom

### LINKED ISSUE #1093: Feature request: Encapsulate styles using Shadow DOM
I'm using Worker DOM to run embedded 3rd party code.

If a third-party insert style like `<style>* { color: red !important }</style>` in the Worker, it can easily affect the host.

Although it was confirmed that encapsulation was possible simply by using an element attached through a shadow root instead of directly using the base element.

Wouldn't this be nice on upstream too? If that's okay, I'll file a PR.

cometkim:
Resolves #1093 

Users can use the Shadow DOM instead of the original, through the `data-shadow-dom` property of the base element.

`data-shadow-dom` attribute should be `open` or `closed`

When Shadow DOM is enabled, it can protect the host styles by isolating style changes that occur inside the Worker.
Uses of `<link>` for external stylesheets inside the Worker are also isolated, so they also need to be moved inside the base element.

---- 

Note:

Using shadow dom may have side effects. I leave some issue I found here.

1. In the React example, event listeners don't work at all. I'm guessing it's a React bug since there's no problem in the same Preact example.

2. Shadow DOM is known to have very little performance penalty, but in high-load apps like DBMon seem to have frame drops quite noticiable actually.
image


kristoferbaxter:
This looks amazing! I'll give it a proper review this week.

src/main-thread/install.ts
```diff
@@ -37,6 +37,13 @@ export function install(
   baseElement: HTMLElement,
   config: InboundWorkerDOMConfiguration,
 ): Promise<ExportedWorker | null> {
+  const mode = baseElement.dataset['shadowDom'];
+  if (mode != null) {
+    const shadowRoot = baseElement.attachShadow({ mode: (mode || 'open') as ShadowRootMode });
```
samouri, line 42: As written this will support anything truthy.
It may be better to require precisely 'open' or 'closed' so that in the future if new modes are added
we don't need to break compat to support

```suggestion
  if (mode === 'open' || mode === 'closed') {
    const shadowRoot = baseElement.attachShadow({ mode });
```

cometkim:
Honestly, I'm not familiar with the environment of doing DOM integration tests here yet, I'll have to spend a bit of time this codebase.

PR #1107, opened 2021-10-28
more descriptive type signature for tuples

cometkim:


src/main-thread/install.ts
```diff
@@ -33,7 +33,7 @@ export function fetchAndInstall(baseElement: HTMLElement, config: InboundWorkerD
  * @param config
  */
 export function install(
-  fetchPromise: Promise<[string, string]>,
+  fetchPromise: Promise<[domScriptContent: string, authorScriptContent: string]>,
```
samouri, line 36:
```suggestion
  fetchPromise: Promise<[workerScript: string, authorScript: string]>,
```

kristoferbaxter (commented):
This is a really good idea. Thank you and @samouri for making the code more readable!

kristoferbaxter (approved):

PR #1132, opened 2022-02-11
Add support for nextElementSibling and previousElementSibling

samouri:
Use case: amp-carousel's buildDom

### DIFF
--- a/src/test/node/nextSibling.test.ts
+++ b/src/test/node/nextSibling.test.ts
@@ -1,14 +1,16 @@
 import anyTest, { TestInterface } from 'ava';
+import { Comment } from '../../worker-thread/dom/Comment';
 import { Element } from '../../worker-thread/dom/Element';
-import { Node } from '../../worker-thread/dom/Node';
+import { Text } from '../../worker-thread/dom/Text';
 import { createTestingDocument } from '../DocumentCreation';
 
 const test = anyTest as TestInterface<{
   node: Element;
   child: Element;
   childTwo: Element;
-  textNode: Node;
-  textNodeTwo: Node;
+  textNode: Text;
+  textNodeTwo: Text;
+  commentNode: Comment;
 }>;
 
 test.beforeEach((t) => {
@@ -20,6 +22,7 @@ test.beforeEach((t) => {
     childTwo: document.createElement('div'),
     textNode: document.createTextNode('Hello'),
     textNodeTwo: document.createTextNode('World'),
+    commentNode: document.createComment('comment'),
   };
 });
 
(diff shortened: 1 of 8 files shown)


src/worker-thread/dom/Element.ts
```diff
@@ -161,6 +159,36 @@ export class Element extends ParentNode {
   // Mixins not implemented
   // Slotable.assignedSlot – https://developer.mozilla.org/en-US/docs/Web/API/Slotable/assignedSlot
 
+  /**
+   * @see https://developer.mozilla.org/en-US/docs/Web/API/Element/previousElementSibling
+   * Returns the Element immediately prior to the specified one in its parent's children list,
+   * or null if the specified element is the first one in the list.
+   */
+  get previousElementSibling() {
```
jridgewell, line 167: This also needs to be implemented on `CharacterData` for `Text` nodes

jridgewell (changes requested):

src/worker-thread/dom/Element.ts
samouri: Done

src/test/node/nextSibling.test.ts
```diff
@@ -38,4 +45,19 @@ test('when a node is the last child of a parent, the next sibling is null', (t)
   node.appendChild(child);
   node.appendChild(childTwo);
   t.is(childTwo.nextSibling, null);
+  t.is(childTwo.nextElementSibling, null);
+});
+
+test('nextElementSibling skips over text nodes', (t) => {
```
rcebulko, line 51: should/does it skip comment nodes?

src/worker-thread/dom/elementSibling.ts
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  for (let i = parentNodes.indexOf(node) + 1; i < parentNodes.length; i++) {
+    let node = parentNodes[i];
+    if (node.nodeType === NodeType.ELEMENT_NODE) {
+      return node as Element;
+    }
+  }
+  return null;
+}
```
rcebulko, line 33: Not sure if/how much bundle size matters for work-dom, but this could save some by sharing the logic (at the cost of some readability)

```suggestion
function getElemementSibling(startNode: Node, step: 1 | -1): Element | null {
  let parentNodes = startNode.parentNode?.childNodes;
  if (!parentNodes) {
    return null;
  }

  let i = parentNodes.indexOf(startNode) + step;
  let node;
  while (node = parentNodes[i]) {
    if (node.nodeType === NodeType.ELEMENT_NODE) {
      return node as Element;
    }
    i += step;
  }
  return null;
}

export function getPreviousElementSibling(node: Node): Element | null {
  return getElemementSibling(node, -1);
}

export function getNextElementSibling(node: Node): Element | null {
  return getElemementSibling(node, 1);
}
```

rcebulko (approved):

src/test/node/nextSibling.test.ts
samouri: it should/does! can update the test

jridgewell (approved):

src/worker-thread/dom/elementSibling.ts
samouri: code size is most definitely a concern. but there are much bigger fish to fry IMO before we hurt readability

samouri:
Thank you for the reviews @jridgewell and @rcebulko!! 🎉

PR #1136, opened 2022-02-22
Storage: fix clear and removeItem

samouri:
**summary**
Fixes: https://github.com/ampproject/amphtml/issues/37739

The mutations array usually contains an index into StringContext for
value of key/value in the operation. Historically, "delete" was
modeled as 0th index into the array. Unfortunately, this didn't make
sense if the first op was genuinely referring to the 0th string
value (See #1035).

While #1035 fixed the first get/set operation, it also broke  delete operations!
This PR now models the delete operation as a -1 in the unsigned mutation array, corresponding to 2^16-1.

The obvious flaw would be if we ever had that many mutations being transferred at once.
From what I can tell, thats more of a theoretical issue than a real one.

### DIFF
--- a/src/transfer/Messages.ts
+++ b/src/transfer/Messages.ts
@@ -107,3 +107,7 @@ export const enum ResolveOrReject {
   RESOLVE = 1,
   REJECT = 2,
 }
+
+// Mutations are are modeled as Uint16Array.
+// Deletion is specified as a -1, which when placed into the array becomes 2^16-1.
+export const DELETION_INDEX = 2 ** 16 - 1;
--- a/src/transfer/Messages.ts
+++ b/src/transfer/Messages.ts
@@ -107,7 +107,3 @@ export const enum ResolveOrReject {
   RESOLVE = 1,
   REJECT = 2,
 }
-
-// Mutations are are modeled as Uint16Array.
-// Deletion is specified as a -1, which when placed into the array becomes 2^16-1.
-export const DELETION_INDEX = 2 ** 16 - 1;
--- a/src/worker-thread/Storage.ts
+++ b/src/worker-thread/Storage.ts
@@ -62,7 +62,7 @@ export function createStorage(document: Document | DocumentStub, location: Stora
         TransferrableMutationType.STORAGE,
         GetOrSet.SET,
         location,
-        store(key),
+        store(key) + 1,
         0, // value == 0 represents deletion.
       ]);
     },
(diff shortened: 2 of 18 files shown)


src/transfer/Messages.ts
```diff
@@ -107,3 +107,7 @@ export const enum ResolveOrReject {
   RESOLVE = 1,
   REJECT = 2,
 }
+
+// Mutations are are modeled as Uint16Array.
+// Deletion is specified as a -1, which when placed into the array becomes 2^16-1.
```
jridgewell, line 112: Why are we setting them to `-1` in the first place? Wouldn't `0` work?

samouri: 0 is overloaded. It can also mean the 0th index into the Strings array

jridgewell: Just define it as index as the `+ 1` of the strings index? Then 0 is impossible.

samouri:
*update / background**
The main-thread code holds on to an array of strings, called `StringContext`. This way the worker-thread only has to send any string over the postMessage interface once. It can then refer to that string via its index into the array. When a value might sometimes exist and might be null, the worker-thread historically had passed in the `0th` index. Unfortunately, that index wasn't actually reserved, so it could also have a string in it & be necessary during hydration.

This PR reserves the 0th slot in main-thread StringContext, and starts the count of the worker-thread from 1.

src/transfer/Messages.ts
samouri: Always sending over +1 from the worker seems messier on both ends than reserving the 0th index. WDYT

samouri:
After a discussion with @jridgewell, we agreed that while reserving 0th slot would be simpler, it may have  unexpected ramifications to the codebase vs isolating to a single processor. Using the +1 in worker, -1 in main-thread approach to isolate the change.

Can do the 0 reservation in a future PR

src/worker-thread/Storage.ts
```diff
@@ -62,7 +62,7 @@ export function createStorage(document: Document | DocumentStub, location: Stora
         TransferrableMutationType.STORAGE,
         GetOrSet.SET,
         location,
-        store(key),
+        store(key) + 1,
```
jridgewell, line 65: There's a `store(key)` and `store(stringValue)` on Line 54 above that need a `+ 1`.

jridgewell (changes requested):

src/worker-thread/Storage.ts
samouri: You're right (of course)! Both the get and set flow use the same logic for determining key/value. I need to figure out why no test cases broke.

samouri: Ah. Because there are no integration tests. Just updated the unit tests -- but this is an obvious gap in worker-dom tests.

jridgewell (approved):

PR #1148, opened 2023-06-30
Unsubscription/removing "on" listener must not cause an error

ddenisyuk:
Removing of event listener with in a way like: `element.onclick = null;`, stores `null` as a handler.

### DIFF
--- a/src/worker-thread/dom/HTMLElement.ts
+++ b/src/worker-thread/dom/HTMLElement.ts
@@ -22,7 +22,9 @@ export const appendGlobalEventProperties = (keys: Array<string>): void => {
         if (stored) {
           this.removeEventListener(normalizedKey, stored);
         }
-        this.addEventListener(normalizedKey, value);
+        if (value instanceof Function) {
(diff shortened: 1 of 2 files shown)


olitvinenko (approved):

src/worker-thread/dom/HTMLElement.ts
```diff
@@ -22,7 +22,9 @@ export const appendGlobalEventProperties = (keys: Array<string>): void => {
         if (stored) {
           this.removeEventListener(normalizedKey, stored);
         }
-        this.addEventListener(normalizedKey, value);
+        if (typeof value === 'function') {
```
powerivq, line 25: Can we do `value instanceof Function`?

powerivq (changes requested):
Thanks for the contribution. LGTM otherwise.

src/worker-thread/dom/HTMLElement.ts
ddenisyuk: sure, no problem

powerivq (approved):

Issue #1031, opened 2021-03-17
Possiblity to use this with Webpack@5.x worker loading?


Slapbox:
I'm wondering if there's any way to use this package without needing to assign our worker file to a `<div />` to be "upgraded."

The problem is that our worker thread is being bundled via Webpack at the same time as the rest of our application and the output page is created using an HTML template. I'm not sure how it would be possible to assign the worker scripts to a div in this setup. Is there any way we can use the library in conjunction with the way Webpack bundles workers?

For reference, our code for creating our worker threads right now looks like this:

```js
new Worker(new URL('./heavyLifterWorker', import.meta.url), {
  name: 'heavyLifter',
}),
```

and then Webpack is automatically bundling the worker chunk and injecting it into the page without any input from us. Is there any way to leverage Worker-DOM together with the built-in Webpack worker loading functionality?

samouri:
Hey @Slapbox,

It doesn't need to be a `div` specifically, but in order to use this package you need to assign a worker to a specific DOM node. In your template you'd need to specify the filepath of the generated worker bundle.

Blame4:
@samouri, and how to assign a node to worker with worker-dom

PR #1194, opened 2024-04-02
Es modules import

w05163:
https://github.com/ampproject/worker-dom/issues/1031
support work with webpack worker


w05163:
please close this，i will commit other one with correct user

